PR #29, opened 2014-12-08
Use ifquery for finding the interfaces

vStone:
When using interfaces.d/<foobar>, interfaces are not picked up. Let ifquery figure out the format of the interfaces files for us.


blp:
> It's a bug if you define interfaces in /etc/interfaces.d/* or any other included file.

Thanks, I cherry-picked this to branch-2.3 then.

PR #31, opened 2014-12-16
Export OVS datapath performance counters via sFlow

sflow:
The OVS cache hit/miss counters and memory/CPU usage statistics have been identified
as important metrics when managing large deployments. This patch allows them
to be pushed periodically as part of the sFlow feed,  and represents a more
efficient and scalable alternative to polling via ovs-dpctl(1).

Signed-off-by: Neil McKee <neil.mckee@inmon.com>


blp:
Applied (with some minor change, see mailing list).  Thanks a lot!

PR #33, opened 2015-02-11
Preserve IPv6 link-local address on interface save

raorn:
If IPv6 link-local address is removed from interface, it is unable to
receive any IPv6 packets, including Route Advertisements.

In save_interface only skip IPv4 "scope link" addresses.


blp:
Applied.

PR #36, opened 2015-03-01
ovs-docker: check if port is already attached for container/interface

giangi:
Reuse code for determining attached port to prevent ovs-docker to proceed if a
port for the given container and interface is already attached.

### DIFF
--- a/utilities/ovs-docker
+++ b/utilities/ovs-docker
@@ -73,6 +73,14 @@ add_port () {
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


blp:
I posted this patch to the ovs-dev mailing list because it is more likely to get reviewed there. The post is visible at: http://openvswitch.org/pipermail/dev/2015-February/051802.html

shettyg:
Applied as commit 709344386fd84b

PR #42, opened 2015-04-29
Add MTU parameter for the network docker interface

apoulain:
Signed-off-by: Aurélien Poulain <aurepoulain@viacesi.fr>

I would like to submit a change because of a problem I encounter.
Problem:
I use a GRE tunnel between several hosts (each host have containers). I can ping containers from one host to another. When I start an iperf, the TCP connection is established, but the MTU on the container blocks the data because the MTU is too high. Indeed, the MTU appears to be 1500 for the container’s interface while it is limited by the GRE tunnel. The GRE tunnels need to encapsulate several protocols:
-Ethernet: 14 bytes
- IPv4: 20 bytes
- GRE 4 bytes
The total overhead is 38 bytes on the physical interface, so the resulting PDU size is 1462 bytes (out of a regular 1500 MTU based interface).

Solution:
I would like to add the possibility to set the MTU parameter for the created network container interface, because it’s not possible to change the MTU inside the container, we have this error:
“Unable to resolve host ######
RTNETLINK answers : Operation not permitted”
We cannot use the network namespace after the ovs-docker script, we have this error:
“cannot open network namespace “#####”: No such file or directory”

In the submitted changes, I have added an MTU argument for the add_port command and I have added a condition to check if the MTU value is present to change the MTU for the interface to the new specified value.

### DIFF
--- a/utilities/ovs-docker
+++ b/utilities/ovs-docker
@@ -196,14 +196,15 @@ ${UTIL}: Performs integration of Open vSwitch with Docker.
 usage: ${UTIL} COMMAND
 
 Commands:
-  add-port BRIDGE INTERFACE CONTAINER [ADDRESS [GATEWAY]]
+  add-port BRIDGE INTERFACE CONTAINER [ADDRESS [GATEWAY [MTU]]]
                     Adds INTERFACE inside CONTAINER and connects it as a port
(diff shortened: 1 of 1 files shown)


shettyg:
Duplicate of #44

PR #46, opened 2015-05-19
netdev-bsd: include net/bpf.h

dankm:
The documentation says it is required to use bpf ioctls on both
NetBSD and FreeBSD. It causes a compile time failure on FreeBSD 10.

### DIFF
--- a/lib/netdev-bsd.c
+++ b/lib/netdev-bsd.c
@@ -26,6 +26,7 @@
 #include <sys/ioctl.h>
 #include <sys/socket.h>
 #include <sys/sockio.h>
+#include <net/bpf.h>
 #include <ifaddrs.h>
 #include <pcap/pcap.h>
 #include <net/if.h>


dankm:
Looks like that failure was unrelated. This code doesn't even run on Linux.

blp:
I think that this is the same as #47, so I'm closing this for now (as I already responded in #47).  If I'm wrong, and they're different, please do reopen this.

dankm:
I'm okay with closing this pull request. The only difference between this one and #47 is this is against master and #47 is against branch-2.3.

blp:
OK, that's fine, thanks!

There's no need to submit one request per branch. We prefer to take changes on matter initially, then backport as necessary.

blp:
Master I mean, not matter.

dankm:
That makes sense, either way, the bpf branch and the bpf-2.3 branch have both been updated with a SOB.

PR #47, opened 2015-05-19
netdev-bsd: include net/bpf.h

dankm:
The documentation says it is required to use bpf ioctls on both
NetBSD and FreeBSD. It causes a compile time failure on FreeBSD 10.


blp:
This seems reasonable.  Can you provide a Signed-off-by for this patch?  The syntax and, more importantly, the meaning of Signed-off-by is explained in CONTRIBUTING.md.

blp:
Thanks for the S-o-b and the patch!  I applied this to master and branch-2.3.

PR #48, opened 2015-05-23
datapath: do not add vlan_hwaccel_push_inside() for kernel >= 3.18.13

hauke:
The vlan_hwaccel_push_inside() function was backported in this commit
to kernel 3.18.13:

commit a67e2e88342accd49587d9bad72f6dabd7673f7c
Author: Jiri Pirko <jiri@resnulli.us>
Date:   Wed Nov 19 14:04:59 2014 +0100

    vlan: introduce *vlan_hwaccel_push_inside helpers

    [ Upstream commit 5968250c868ceee680aa77395b24e6ddcae17d36 ]

Without this patch compilation breaks on kernel >= 3.18.13

Signed-off-by: Hauke Mehrtens <hauke@hauke-m.de>

### DIFF
--- a/datapath/linux/compat/include/linux/if_vlan.h
+++ b/datapath/linux/compat/include/linux/if_vlan.h
@@ -52,7 +52,7 @@ static inline struct sk_buff *rpl_vlan_insert_tag_set_proto(struct sk_buff *skb,
 }
 #endif
 
-#if LINUX_VERSION_CODE < KERNEL_VERSION(3,19,0)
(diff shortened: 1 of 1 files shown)


blp:
I'm not certain that any of the OVS datapath developers track pull requests, so I posted this to ovs-dev: http://openvswitch.org/pipermail/dev/2015-May/055563.html

blp:
Jesse commented in email: "I think it's probably better to use a configure check here. Given that this was already backported to -stable, it seems like there is a pretty high probability that it could show up in earlier kernels too."

On that basis, I recommend rewriting this as a configure check.

PR #52, opened 2015-06-25
Fixes in python IDL & JSONRPC probes

sumgarg:
Please see commit comments for details


blp:
I broke the first commit into two, one for each bug that it fixes, and pushed all three commits to master, branch-2.4, and branch-2.3.  Thank you!

PR #53, opened 2015-06-29
fix wrong cmd for create dpdkvhostuser

Grim-lock:
Signed-off-by: Wei li <liw@dtdream.com>


blp:
Thanks, I applied this patch to master and branch-2.4.  I'll also send it to the ovs-dev mailing list so that others notice it.

PR #60, opened 2015-07-15
ovs-appctl: register unixctl commands in the beginning

hanxueluo:
Some commands of ovs-appctl were lazily registered when first
bridg or bfd was added. Before that, calling these commands raised a
error("xxx is not a valid command"). The problem commangs included
"bfd/...", "upcall/...","dpif/...","fdb/..." and so on.

Fix this by moving the register into the "bridge_init" and
"bridge_init_ofproto". All commands are registered at the moment
ovs-vswitchd starts.

### DIFF
--- a/lib/bfd.c
+++ b/lib/bfd.c
@@ -329,6 +329,16 @@ bfd_get_status(const struct bfd *bfd, struct smap *smap)
     ovs_mutex_unlock(&mutex);
 }
 
+void
+bfd_init(void)
+{
+    unixctl_command_register("bfd/show", "[interface]", 0, 1,
+                             bfd_unixctl_show, NULL);
+    unixctl_command_register("bfd/set-forwarding",
+                             "[interface] normal|false|true", 1, 2,
+                             bfd_unixctl_set_forwarding_override, NULL);
+}
+
 /* Initializes, destroys, or reconfigures the BFD session 'bfd' (named 'name'),
  * according to the database configuration contained in 'cfg'.  Takes ownership
  * of 'bfd', which may be NULL.  Returns a BFD object which may be used as a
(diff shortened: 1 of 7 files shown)


hanxueluo:
The way to reproduce bug:
1. delete all bridges
2. restart ovs
3. call command: "ovs-appctl dpif/show" or "ovs-appctl fdb/show" or "ovs-appctl upcall/show" ....
The cmds reports error "XXX is not a valid com".

yew011:
I'm not sure this is a 'bug'.

Since these appctl sub-commands either require 'bridge' to be provided as argument or query the 'bridge' related property (e.g. bfd/cfm can only be enabled on iface, and iface is part of a bridge)

Could you explain more on how you bothered by the error message?

Thanks,
Alex Wang,

hanxueluo:
Case 1:
1. start ovs with some bridges in ovsdb, **then delete all bridges**
2.1 execute "ovs-appctl dpif/show" 
      ==> it outputs nothing. 
2.2 execute "ovs-appctl fdb/show" 
      ==> it outputs:*"fdb/show" command requires at least 1 arguments\rovs-appctl: ovs-vswitchd: server returned an error*
2.3 execute "ovs-appctl fdb/show abc" 
    ==> it outputs:*no such bridge\r ovs-appctl: ovs-vswitchd: server returned an error*

Case 2:
1. start ovs **without any bridge in ovsdb**
2.1 execute "ovs-appctl dpif/show" 
2.2 execute "ovs-appctl fdb/show" 
2.3 execute "ovs-appctl fdb/show abc" 
    ==> they all output: *"xxxx" is not a valid command\r   ovs-appctl: ovs-vswitchd: server returned an error*

The outputs in case 1 are all reasonable. Because we do have no bridge and pass wrong argument. However, the output of case 2 report incorrect error message. 

I think there should be no difference for the cmds between such 2 cases. 
In addition, "cfm" has a different implement from "bfd". it registers its commands in 
bridge_init() -> cfm_init().

blp:
I think that giving different errors in case 1 and case 2 is likely to surprise users.
I applied this patch to master and branch-2.4
Thank you!

PR #62, opened 2015-08-06
Fix typo

Grim-lock:
Fix typo


blp:
It looks like there's only one commit that isn't already in master here, the "README-native-tunneling.md : Fix typo." commit.  Can you resubmit that separately?  Thanks.

PR #63, opened 2015-08-06
Add VLAN support for "add_port" function

LazGit:
To be used in a one-liner, like pipework from jpetazzo, for containers creation. It's avoid a second step which require CONTAINER_ID again if you use VLAN.

Example : ovs-docker add-port ovs-br0 eth0 $(docker run -d centos:7) --ipaddress=10.0.0.2/24 --gateway=10.0.0.1 --vlan=1010

### DIFF
--- a/utilities/ovs-docker
+++ b/utilities/ovs-docker
@@ -90,6 +90,11 @@ add_port () {
                 MTU=`expr X"$1" : 'X[^=]*=\(.*\)'`
                 shift
                 ;;
+            --vlan=*)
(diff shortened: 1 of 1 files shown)


shettyg:
I am happy to commit this. But I need the following line in the commit message:

Signed-off-by: Ludovic Paret <ludovivc.paret@atos.net>

The meaning of the above line is explained in
https://github.com/openvswitch/ovs/blob/master/CONTRIBUTING.md

blp:
Closing since there has been no response from the original poster since Aug. 6.

PR #65, opened 2015-09-04
Fix rhel build

dotSlashLu:
`rpmbuild` on CentOS 6 is broken for two errors, detailed error messages can be found in each commit message. The first one, `cannot remove <directory>: no such file or directory`, can be securely fixed by patch 1.
But for the second one, `File not found`, I'm not sure the cause of the error, it's weird those binaries/man pages can't be found after the build succeeded. I guess those binaries are no longer being distributed, but the rpm build spec was left out-dated, if this is the case, then the second patch can be securely applied.


blp:
Here is some feedback via IRC.  "blp" is me, "fbl" is Flavio Leitner, a Red Hat employee:

blp: fbl: Any chance you could take a look at these rhel fixes? https://github.com/openvswitch/ovs/pull/65
blp: fbl: I find them a little puzzling because it seems like multiple users would be reporting them.
switchcade: incidentally, that pull request reports travis failures but they seem to be unrelated. I suspect no-one's followed up on travis failures in the last week or so.
blp: switchcade: You're probably right. If I could get travis to email me, I'd check on failures, but it's frustratingly unwilling to do that.
fbl: blp: I am not building rpm on centos6, but the rm command could be fixed to use -f or simply remove the missing files.  In the other patch, there is no reason to comment out a file that doesn't exist (assuming it's not broken), just take them out.
blp: fbl: I guess I don't see why the situation would be different here on RHEL vs. Centos or 6 vs. other versions.
fbl: blp: most people is probably using rhel7/centos7/or fedora14 and above, and then the openvsiwotch-fedora.spec is better.
blp: fbl: I didn't realize that. Thanks.
fbl: which is the one I am using to keep fedora and RHEL 7/OSP

dotSlashLu:
Thanks for the feedback. Are the missing files
- /usr/bin/ovs-parse-backtrace
- /usr/bin/ovs-pcap
- /usr/bin/ovs-tcpundump
- /usr/bin/ovs-vlan-test
- /usr/sbin/ovs-bugtool
- /usr/share/man/man8/ovs-bugtool.8.gz
- /usr/share/openvswitch/bugtool-plugins
- /usr/share/openvswitch/python
- /usr/share/openvswitch/scripts/ovs-bugtool-*
- /usr/share/man/man1/ovs-sim.1.gz

really supposed to be missing? Sounds true for me based on Flavio's words. Just to make sure.  

If it's true, I'll check rhel/openvswitch.spec against openvsiwotch-fedora.spec which is considered to be better and update my PR to remove those lines rather than commenting it out. Is that ok?

blp:
I guess that they are supposed to be missing because openvswitch-fedora.spec marks them as %exclude.  I don't know why they are excluded--looking back in the history, I see that they were excluded from the point that they were introduced.

I guess it might be worth asking on ovs-dev why these files are not packaged.  They are packaged in Debian, except for ovs-sim, which isn't meant to be installed at all.

dotSlashLu:
Also fixed a new error when building kmod rpm. 
I'll submit a patch hours later and see what ovs-dev will say about the missing files.

dotSlashLu:
fixed with f929235bbe71e2abf330eaf93068b9814a6a2d4b and 43b6e7ca4c70c7fc0dfad0de49b1d907f4068289

PR #67, opened 2015-09-10
ofctl: This patch add support for setting the first egress table for egress processing.

nitirohilla:
"ovs-ofctl set-first-egress-table <SWITCH> <table_id>" can be used to set first egress
table.
This patch enhances TABLE_FEATURES_REQUEST message to set the first egress table and
TABLE_FEATURES_REPLY to identify the table which is configured as first egress table.

Signed-off-by: Niti Rohilla <niti.rohilla@tcs.com>


blp:
Please fix the error reported by travis and resubmit.

PR #69, opened 2015-09-11
ofproto: Implement OF1.4 error code for set-async-config

ambikaarora:
This patch adds support for Openflow1.4 error codes for set-async-config.
In this patch, a new error type, OFPET_ASYNC_CONFIG_FAILED is introduced
that enables the switch to properly inform the controller when controller
tries to set invalid mask or unsupported configuration.

Signed-off-by: Ambika Arora <ambika.arora@tcs.com>

### DIFF
--- a/lib/ofp-errors.h
+++ b/lib/ofp-errors.h
@@ -623,6 +623,19 @@ enum ofperr {
     /* ONF1.3(4448), OF1.4+(14,8).  Permissions error. */
     OFPERR_OFPBPC_EPERM,
 
+/* ## -------------------------- ## */
+/* ## OFPET_ASYNC_CONFIG_FAILED  ## */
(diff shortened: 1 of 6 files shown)


blp:
Sorry about the delay in review.  It's easier to keep track of patches that are posted to ovs-dev.

This looks like a step in the right direction, but it also looks difficult to maintain because it adds new and nonobvious place to update (the MAXIMUM_* constants) whenever a new asynchronous message is defined.  Can you think of a way to refactor it so that that becomes harder to get wrong?

blp:
It's been over a month since my comment above.  I don't think I've seen an update since then (please correct me if I'm wrong) so I'm going to close this as it seems like that indicates it's been abandoned.

PR #71, opened 2015-09-14
bond: Fix slb-balance heuristic calculations.

gasmith-nutanix:
The choose_entry_to_migrate() helper tries to determine whether moving a single
MAC hash from one bond slave to another would make a tangible improvement in
balance. The crucial comparison is a delta in the ratios of tx load between
bond slaves: if the delta is greater than a threshold, the move is recommended.

It is important that both ratios are greater than (or less than) 1 for this
delta to make sense; otherwise, a very heavy MAC hash could ping-pong between
bond slaves at every rebalance interval.

### DIFF
--- a/ofproto/bond.c
+++ b/ofproto/bond.c
@@ -1074,8 +1074,18 @@ choose_entry_to_migrate(const struct bond_slave *from, uint64_t to_tx_bytes)
         }
 
         delta = e->tx_bytes;
+        if (from->tx_bytes <= delta) {
+            /* All of the load on 'from' comes from a single MAC hash, so
+             * shifting load away would be pointless. */
+            return NULL;
+        }
         old_ratio = (double)from->tx_bytes / to_tx_bytes;
         new_ratio = (double)(from->tx_bytes - delta) / (to_tx_bytes + delta);
+        if (new_ratio < 1.0) {
+            /* Normalize the new ratio, such that the most heavily loaded slave
+             * is the numerator, in order to assess the improvement. */
+            new_ratio = 1.0 / new_ratio;
+        }
         if (old_ratio - new_ratio > 0.1
             && fabs(new_ratio - 1.0) < fabs(old_ratio - 1.0)) {
             /* We're aiming for an ideal ratio of 1, meaning both the 'from'


azhou-nicira:
Thanks for reporting the issue and working on it.

I believe the first hunk should have been covered by the list_is_short() check above.

The second hunk looks correct. But it makes the logic harder to understand. I believe
this is due to original logic being more complicated than necessary. I proposed the following
patch trying to simplify:

http://openvswitch.org/pipermail/dev/2015-September/060076.html

It should also address the bug you found. Do you mind testing it in your environment?

gasmith-nutanix:
I prefer your proposal to mine. I'll run some tests and get back to you.

Thanks,
Greg

From: Andy Zhou
Reply-To: openvswitch/ovs
Date: Tuesday, September 15, 2015 at 2:22 PM
To: openvswitch/ovs
Cc: Greg Smith
Subject: Re: [ovs] bond: Fix slb-balance heuristic calculations. (#71)


Thanks for reporting the issue and working on it.

I believe the first hunk should have been covered by the list_is_short() check above.

The second hunk looks correct. But it makes the logic harder to understand. I believe
this is due to original logic being more complicated than necessary. I proposed the following
patch trying to simplify:

http://openvswitch.org/pipermail/dev/2015-September/060076.html

It should also address the bug you found. Do you mind testing it in your environment?

azhou-nicira:
Thanks for reporting the issue and working on it.

I believe the first hunk should have been covered by the list_is_short()
check above.

The second hunk looks correct. But it makes the logic harder to understand.
I believe
this is due to original logic being more complicated than necessary. I
proposed the following
patch trying to simplify:

http://openvswitch.org/pipermail/dev/2015-September/060076.html

It should also address the bug you found. Do you mind testing it in your
environment?

gasmith-nutanix:
Tried out the patch, and it looks like everything is working as expected. Thanks again for the fix!

azhou-nicira:
Thanks for testing it.  I will wait for Ben to get chance to look at it
before committing it to master and then back port it some older branches.

blp:
It looks like the issue that this addresses was fixed with commit c460a6a7bc751 (ofproto/bond: simplify rebalancing logic), so I'm closing it.  If I misunderstand, please reopen it or submit a new patch.

PR #74, opened 2015-09-22

shadansari:

This change adds support for:
- Allow change tracking of specific columns. This ensures that changed
   entire table.
   and before ovsdb_idl_run().


blp:
This pull request is weird: it has two commits with different content but identical commit messages.  I guess they should be squashed together.  Can you submit a pull request with just a single squashed commit?  Thanks.

PR #75, opened 2015-09-30

shadansari:

This change adds support for:
- Change tracking of specific columns. This ensures that changed
   entire table.
   and before ovsdb_idl_run().

Example Usage:

{
   idl = ovsdb_idl_create(...);

   /* Track specific columns */
   ovsdb_idl_track_add_column(idl, column);

   /* Or, track all columns */
   ovsdb_idl_track_add_all(idl);

   for (;;) {
     ovsdb_idl_run(idl);
     seqno = ovsdb_idl_get_seqno(idl);

     /* Process only the changed rows in Table FOO */
     FOO_FOR_EACH_TRACKED(row, idl) {
       /* Determine the type of change from the row seqnos */
       if (foo_row_get_seqno(row, OVSDB_IDL_CHANGE_DELETE) > seqno)) {
         /* row is deleted */
       } else if {foo_row_get_seqno(row, OVSDB_IDL_CHANGE_MODIFY) > seqno)) {
         /* row is modified */
       } else if {foo_row_get_seqno(row, OVSDB_IDL_CHANGE_INSERT) > seqno)) {
         /* row is inserted */
       }
     }

     /* All changes processed - clear the change track */
     ovsdb_idl_track_clear(idl);
   }
}

### DIFF
--- a/lib/ovsdb-idl-provider.h
+++ b/lib/ovsdb-idl-provider.h
@@ -36,6 +36,9 @@ struct ovsdb_idl_row {
     unsigned long int *prereqs; /* Bitmap of columns to verify in "old". */
     unsigned long int *written; /* Bitmap of columns from "new" to write. */
     struct hmap_node txn_node;  /* Node in ovsdb_idl_txn's list. */
+
+    unsigned int change_seqno[OVSDB_IDL_CHANGE_MAX];
+    struct ovs_list track_node;
 };
 
 struct ovsdb_idl_column {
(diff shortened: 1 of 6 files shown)


shadansari:
closing to fix commit message garble

ajavier:
Hi Shad,

This change seems to produce a lot of benefits for the OVSDB subscribers.

Although, you are making the IDL memory usage grow, since the IDL data
structure needs to keep the change list (a lot of changes represents more
memory usage). Have you consider to register callbacks to perform the task
instead? Callbacks would be functions provided by the developer and they
could be called -inmediatelly- when the notification is received, instead
of keeping the response in the IDL.

Such implementation might save memory and speed the process up, since the
memory allocation, data storage and memory deallocation consume CPU cycles.

What do you think?

Regards.

shadansari:
Callback's are certainly one way to go. And the python IDL has a notify method (which is not implemented). Maybe this is something you could propose with the python idl.

blp:
I'm against callbacks, at least in C, because they make code flow unpredictable.

PR #78, opened 2015-10-07

shadansari:

This change adds support for:
  - Change tracking of specific columns. This ensures that changed
     entire table.
     and before ovsdb_idl_run().

    ```
    /* Example Usage */
    {
       idl = ovsdb_idl_create(...);
       /* Track specific columns */
       ovsdb_idl_track_add_column(idl, column);
       /* Or, track all columns */
       ovsdb_idl_track_add_all(idl);
       for (;;) {
         ovsdb_idl_run(idl);
         seqno = ovsdb_idl_get_seqno(idl);
         /* Process only the changed rows in Table FOO */
         FOO_FOR_EACH_TRACKED(row, idl) {
           /* Determine the type of change from the row seqnos */
           if (foo_row_get_seqno(row, OVSDB_IDL_CHANGE_DELETE) > seqno)) {
             /* row is deleted */
           } else if {foo_row_get_seqno(row, OVSDB_IDL_CHANGE_MODIFY) > seqno))
             /* row is modified */
           } else if {foo_row_get_seqno(row, OVSDB_IDL_CHANGE_INSERT) > seqno))
             /* row is inserted */
           }
         }
         /* All changes processed - clear the change track */
         ovsdb_idl_track_clear(idl);
       }
    }
    ```

### DIFF
--- a/lib/ovsdb-idl-provider.h
+++ b/lib/ovsdb-idl-provider.h
@@ -36,6 +36,9 @@ struct ovsdb_idl_row {
     unsigned long int *prereqs; /* Bitmap of columns to verify in "old". */
     unsigned long int *written; /* Bitmap of columns from "new" to write. */
     struct hmap_node txn_node;  /* Node in ovsdb_idl_txn's list. */
+
+    unsigned int change_seqno[OVSDB_IDL_CHANGE_MAX];
+    struct ovs_list track_node;
 };
 
 struct ovsdb_idl_column {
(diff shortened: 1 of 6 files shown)


blp:
Thank you for the patch!  I've started a discussion of it on ovs-dev: http://openvswitch.org/pipermail/dev/2015-October/061163.html

I think that it is likely that a new version will be required, so I'm closing this pull request for now.  I hope that later a new version will be posted.

PR #81, opened 2015-10-22
Add ovsdb_port attribute

sspans:
The hardcoded ovsdb port causes problems when hooking up xenserver to different SDN stacks.
Changing this to a variable at the start of the script makes it easier to update this when needed (using chef/puppet/etc)


blp:
This seems reasonable to me.  Can you provide a Signed-off-by: line?

sspans:
Added, as requested.

blp:
I cherry-picked this to master. Thank you!

PR #83, opened 2015-10-22
ovsdb-idl: Support for readonly columns that are fetched on-demand

shadansari:
There is currently no mechanism in IDL to fetch specific column values
on-demand without having to register them for monitoring. In the case
where the column represent a frequently changing entity (e.g. counter),
and the reads are relatively infrequent (e.g. CLI client), there is a
significant overhead in replication.

This patch adds support in the Python IDL to register a subset of the
columns of a table as "readonly". Readonly columns are not replicated.
Users may "fetch" the readonly columns of a row on-demand. Once fetched,
the columns are not updated until the next fetch by the user. Writes by
the user to readonly columns does not change the value (both locally or
on the server).

The two main user visible changes in this patch are:
  - The SchemaHelper.register_columns() method now takes an optionaly
    argument to specify the subset of readonly column(s)
  - A new Row.fetch(columns) method to fetch values of readonly columns(s)

Usage:
------

    # Schema file includes all columns, including readonly
    schema_helper = ovs.db.idl.SchemaHelper(schema_file)

    # Register interest in columns with 'r' and 's' as readonly
    schema_helper.register_columns("simple", [i, r, s], [r, s])

    # Create Idl and jsonrpc, and wait for update, as usual
    ...

    # Fetch value of column 'r' for a specific row
    row.fetch('r')
    txn.commit_block()

    print row.r
    print getattr(row, 'r')

    # Writing to readonly column has no effect (locally or on server)
    row.r = 3
    print row.r     # prints fetched value not 3

### DIFF
--- a/python/ovs/db/idl.py
+++ b/python/ovs/db/idl.py
@@ -107,6 +107,7 @@ class Idl(object):
         schema = schema.get_idl_schema()
 
         self.tables = schema.tables
+        self.readonly = schema.readonly
         self._db = schema
         self._session = ovs.jsonrpc.Session.open(remote)
         self._monitor_request_id = None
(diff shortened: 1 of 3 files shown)


blp:
I cherry-picked this to master.  Thank you!

PR #86, opened 2015-11-06
fix bug :Memory leak

xushengping:
This causes a memory leak bug


blp:
Thank you for the report, and the patch! This is a duplicate of #87, I believe, which has been fixed, so I'm closing it.

PR #87, opened 2015-11-10
fix bug

xushengping:
here can lead to memory leaks


blp:
Thanks for the report.  I think that we in fact need to round up instead of using <=.  I applied a patch that does that, crediting this pull request:

PR #88, opened 2015-11-11
modify the way to create config_tmpfile

vanpire110:
Glibc will try the path prefix p_tmpdir defined in <stdio.h>, and if that fails the directory /tmp, but p_tmpdir which is "/dev/shm/tmp" may fail again. So we need a safe way to create config_tmpfile

### DIFF
--- a/ovsdb/ovsdb-server.c
+++ b/ovsdb/ovsdb-server.c
@@ -228,10 +228,14 @@ main(int argc, char *argv[])
      * configuration to it.  When --monitor is used, this preserves the effects
      * of ovs-appctl commands such as ovsdb-server/add-remote (which saves the
      * new configuration) across crashes. */
-    config_tmpfile = tmpfile();
+
+    /* config_tmpfile = tmpfile();  */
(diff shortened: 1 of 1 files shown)


blp:
This needs more explanation.  Why doesn't tmpfile() work on your system?

vanpire110:
My system is embedded Linux(windriver linux 6.0) , the  file system is modified. In my system, both   /tmp dir and /dev/shm/tmp dir do not exist. So I need an existed dir to create this tmp file for ovs. And I can confirm dir that ovs_rundir() generated do exist.

xushengping:
On same systems,temporary file directory may not have write permission.

russellb:
I don't think we shouldn't be creating temp files in an arbitrary location like this.  The issue seems to be that you should fix your system such that tmpfile() works properly by having a writable temp file location.

blp:
If temporary files do not have write permission, then your system is broken and should be fixed.  The ANSI C standard clearly says that tmpfile() returns a file that is opened for reading and writing.

blp:
Your system is broken.  You should fix it, rather than fixing every program you run on the system.  In the long run, this will be easier.

vanpire110:
Thank  you for your advice.

PR #92, opened 2015-12-02
ovn-northd: The traffic form gateway port not need to be followed

Grim-lock:
Signed-off-by: l0310 <liw@dtdream.com>

In openstack networking-ovn, about the traffic between two vms in different physical machines and different subnet, the way of ingress and egress traffic of one vm is different. The ingress is not followed by gateway port, so egress will be drop.

### DIFF
--- a/ovn/northd/ovn-northd.c
+++ b/ovn/northd/ovn-northd.c
@@ -964,9 +964,11 @@ has_stateful_acl(struct ovn_datapath *od)
 }
 
 static void
-build_acls(struct ovn_datapath *od, struct hmap *lflows)
+build_acls(struct ovn_datapath *od, struct hmap *lflows, struct hmap *ports)
 {
     bool has_stateful = has_stateful_acl(od);
+    struct ovn_port *op;
+    struct ds matchIn, matchOut;
 
     /* Ingress and Egress Pre-ACL Table (Priority 0): Packets are
      * allowed by default. */
@@ -983,6 +985,30 @@ build_acls(struct ovn_datapath *od, struct hmap *lflows)
      * send all IP packets through the conntrack action, which handles
      * defragmentation, in order to match L4 headers. */
     if (has_stateful) {
+        HMAP_FOR_EACH (op, key_node, ports) {
+            if (op->od == od && !strcmp(op->nbs->type, "router")) {
+                /* Can't use ct() for router ports. Consider the following configuration:
+                lp1(10.0.0.2) on hostA--ls1--lr0--ls2--lp2(10.0.1.2) on hostB,
+                For a ping from lp1 to lp2, First, the response will go through ct()
+                with a zone for lp2 in the ls2 ingress pipeline on hostB.
+                That ct zone knows about this connection. Next, it goes through ct()
+                with the zone for the router port in the egress pipeline of ls2 on hostB.
+                This zone does not know about the connection, as the icmp request
+                went through the logical router on hostA, not hostB. This would only work
+                with distributed conntrack state across all chassis. */
+
+                ds_init(&matchIn);
+                ds_init(&matchOut);
+                ds_put_format(&matchIn, "ip && inport == %s", op->json_key);
+                ds_put_format(&matchOut, "ip && outport == %s", op->json_key);
+                ovn_lflow_add(lflows, od, S_SWITCH_IN_PRE_ACL, 110, ds_cstr(&matchIn), "next;");
+                ovn_lflow_add(lflows, od, S_SWITCH_OUT_PRE_ACL, 110, ds_cstr(&matchOut), "next;");
+
+                ds_destroy(&matchIn);
+                ds_destroy(&matchOut);
+            }
+        }
+
         /* Ingress and Egress Pre-ACL Table (Priority 100).
          *
          * Regardless of whether the ACL is "from-lport" or "to-lport",
@@ -1100,7 +1126,7 @@ build_lswitch_flows(struct hmap *datapaths, struct hmap *ports,
             continue;
         }
 
-        build_acls(od, lflows);
+        build_acls(od, lflows, ports);
     }
 
     /* Logical switch ingress table 0: Admission control framework (priority
--- a/ovn/ovn-nb.xml
+++ b/ovn/ovn-nb.xml
@@ -383,6 +383,11 @@
         restrictive policy, it is important to remember to allow flows
         such as ARP and IPv6 neighbor discovery packets.
       </p>
+
+      <p>
+        Can't create an ACL matching on port with type=router because
+        we do not use conntrack on it.
+      </p>
     </column>
 
     <column name="action">


blp:
I'm having trouble understanding the commit message.  Perhaps you could give an example?  Thanks!

russellb:
I *think* I understand what you're saying ... that in the case of 2 ports on 2 logical switches connected by a logical router with 1 port on each host, the path through conntrack is not symmetric and leads to packets getting dropped when we don't want them to be.

I'm setting this up with a 2-node OpenStack environment to take a closer look.

Grim-lock:
Bryant is right

topo:
vm1(1.1.1.2/24)--lswitch--(1.1.1.1)lrouter(2.2.2.1)--lswitch--vm2(2.2.2.2/24)

vm1 and vm2 is on2 different host, vm1 on host1  vm2 on host2

When ping from vm1 to vm2 , the result of ovs-dpctl dump-flows on host2:

ingress:
recirc_id(0),tunnel(tun_id=0x3,src=10.160.1.122,dst=10.160.1.80,ttl=64,geneve({class=0xffff,type=0,len=4,0x20003}),flags(-df-csum+key)),in_port(2),skb_mark(0),eth_type(0x0800),ipv4(frag=no), 
packets:23, bytes:2254, used:0.314s, 
actions:ct(zone=1,mark=0/0,label=0/0),recirc(0x6)
recirc_id(0x6),tunnel(tun_id=0x3,src=10.160.1.122,dst=10.160.1.80,ttl=64,geneve({class=0xffff,type=0,len=4}),flags(-df-csum+key)),in_port(2),skb_mark(0),ct_state(-new+est-rel-inv+trk),eth(dst=fa:16:3e:a5:d0:28),eth_type(0x0800),ipv4(frag=no), 
packets:22, bytes:2156, used:0.314s, actions:3

egress
recirc_id(0),in_port(3),eth(src=fa:16:3e:a5:d0:28),eth_type(0x0800),ipv4(frag=no), 
packets:23, bytes:2254, used:0.313s, 
actions:ct(zone=1,mark=0/0,label=0/0),recirc(0x7)
recirc_id(0x7),in_port(3),ct_state(-new+est-rel-inv+trk),eth(dst=fa:16:3e:e6:6d:03),eth_type(0x0800),ipv4(frag=no), 
packets:23, bytes:2254, used:0.313s, 
actions:ct(zone=1,mark=0/0,label=0/0),recirc(0x8)
recirc_id(0x8),in_port(3),ct_state(-new+est-rel-inv+trk),eth(src=fa:16:3e:a5:d0:28,dst=fa:16:3e:e6:6d:03),eth_type(0x0800),ipv4(src=2.2.2.2/255.255.255.254,dst=1.1.1.3,ttl=64,frag=no), 
packets:23, bytes:2254, used:0.313s, actions:set(eth(bad key length 24, 
expected 12)(fa 16 3e 8f 97 13 fa 16 3e de 5f 6c 20 00 0b 00 1c 00 07 00 
02 02 02 
02)),set(ipv4(src=2.2.2.2/255.255.255.254,dst=1.1.1.3,ttl=63)),_ct(mark=0/0,label=0/0)_,recirc(0x9)
recirc_id(0x9),in_port(3),ct_state(-new-est-rel+inv+trk),eth_type(0x0800),ipv4(frag=no), 
packets:23, bytes:2254, used:0.313s,actions:drop

_ct(mark=0/0,label=0/0)_ not in the ingress way

ovs-ofctl dump-flows br-int:
table=0, in_port=7 
actions=load:0x3->OXM_OF_METADATA[],load:0x2->NXM_NX_REG6[],resubmit(,16) port7 
is gateway port, reg5 is not set
table=0, in_port=8 
actions=load:0x1->NXM_NX_REG5[],load:0x3->OXM_OF_METADATA[],load:0x3->NXM_NX_REG6[],resubmit(,16)

table=17, priority=100,ip,metadata=0x3 
actions=ct(table=18,zone=NXM_NX_REG5[0..15])

table 17 can not distinguish the traffic between port7 and port8, so 
dpflow has _ct(mark=0/0,label=0/0)_


在 2015/12/3 4:22, Russell Bryant 写道:

russellb:
Sorry for the delay, but I finally got back to looking at this.  I can see the issue now.  I'll try to rephrase it:

Consider the following configuration: lp1 (10.0.0.2) on hostA ↔ ls1 ↔  lr0 ↔ ls2 ↔ lp2 (10.0.1.2) on hostB

First, a review of how we use conntrack zones.  We use the ct() action in both the ingress and egress pipelines for logical switches.  In the ingress pipeline, the conntrack zone is based on the source logical port.  In the egress pipeline, the conntrack zone is based on the destination logical port.  The root of this problem is that logical router ports exist on multiple chassis.

For a ping from lp1 to lp2, the problem occurs on the return path.  First, the response will go through ct() with a zone for lp2 in the ls2 ingress pipeline on hostB.  That ct zone knows about this connection.  Next, it goes through ct() for the zone for the router port in the egress pipeline of ls2 on hostB.  This zone does *not* know about the connection, as the icmp request went through the logical router on hostA, not hostB.  The result is that we can't use ct() for router ports.  This would only work with distributed conntrack state across all chassis.

In this patch, I see that router ports are only excluded for the ingress pipeline.  Is there a reason why the same thing isn't done for the egress pipeline, as well?

This change also limits what you can do with ACLs, so we should probably document that somewhere.  Specifically, you won't be able to create an ACL matching on inport or outport of a logical router port.

Grim-lock:
CT in the router egress pipeline seems not to result any wrong, so I did not delete it.
If no impact to other thing, I think that should delete it in both direction of lrouter until the distributed CT is finally achieved.
or is there another way to fix that?

russellb:
I would just remove it from both directions.  I don't think it adds any value on the egress side if it has been removed on the ingress side.

Grim-lock:
Already remove it from both directions. please check it. :)

ovn/northd/ovn-northd.c
```diff
@@ -983,14 +985,36 @@ build_acls(struct ovn_datapath *od, struct hmap *lflows)
...
+             *
+             * Regardless of whether the ACL is "from-lport" or "to-lport",
+             * we need rules in both the ingress and egress table, because
+             * the return traffic needs to be followed. */
+            ovn_lflow_add(lflows, od, S_SWITCH_IN_PRE_ACL, 100, ds_cstr(&matchIn), "ct_next;");
+            ovn_lflow_add(lflows, od, S_SWITCH_OUT_PRE_ACL, 100, ds_cstr(&matchOut), "ct_next;");
+            
+            ds_destroy(&matchIn);
+            ds_destroy(&matchOut);
+        }
```
russellb: Here's an alternative that would result in a lot less flows, I think.

You could still create the original flows at priority 100 that are removed in this patch.  Then, in the loop, create a new flow at a higher priority when the type *is* router, match on IP and the router port ID, and set the action to "next" instead of "ct_next".  

That way we only have new flows for each router port, which there should be less of.

russellb:
In addition to the inline comment, I think we need to update documentation to say that you can't use port IDs in ACLs for ports with type=router.  I suppose the best place to put that would be the docs for the "match" column of the "ACL" table in ovn/ovn-nb.xml.

blp:
> This change also limits what you can do with ACLs, so we should probably document that somewhere. Specifically, you won't be able to create an ACL matching on inport or outport of a logical router port.

Is that an important limitation?

russellb:
> Is that an important limitation?

I don't think so.  I just figured it was worth documenting explicitly.

blp:
> > Is that an important limitation?

> I don't think so.

Then I guess we should commit this, if you're satisfied with it.

russellb:
On 12/23/2015 04:32 PM, Ben Pfaff wrote:
>         Is that an important limitation?
> 
>     I don't think so.
> 
> Then I guess we should commit this, if you're satisfied with it.

Yes, I will review again and commit it.  I was waiting on the doc
update, and it looks like that's done now.  I wish github would provide
a notification when a pull request is updated with a new revision of a
patch!

-- 
Russell Bryant

blp:
On Mon, Jan 04, 2016 at 10:37:35AM -0800, Russell Bryant wrote:
> On 12/23/2015 04:32 PM, Ben Pfaff wrote:
> >         Is that an important limitation?
> > 
> >     I don't think so.
> > 
> > Then I guess we should commit this, if you're satisfied with it.
> 
> Yes, I will review again and commit it.  I was waiting on the doc
> update, and it looks like that's done now.  I wish github would provide
> a notification when a pull request is updated with a new revision of a
> patch!

I'm surprised it doesn't; that sounds like a basic feature.

russellb:
Thanks for the patch!  I applied this to master and branch-2.5.

PR #93, opened 2015-12-04
ovn-nbctl:add db commands help

Grim-lock:
Signed-off-by: l0310 <liw@dtdream.com>


russellb:
Can you update the man page, as well?

Thanks!

Grim-lock:
I finished according to your idea, but there is a place not good.
In db-ctl-base.man, it uses a var instead of command which to be man,  e.g. command within the same \fB\*(PN\fR invocation. I just write ovn-nbctl in db-ctl-base.xml, it does not change, if ovn-sbctl include him.
How can I do this in db-ctl-base.xml?

russellb:
I see.  I'm not sure the best way to handle that.  I know we could come up with something, but the easiest thing is to just remove it.  It doesn't seem to harm the content to replace "<code>ovn-nbctl</code>" with "command", except the last case which would be "the command".  That's my suggestion unless someone else feels strongly that we need to continue to refer to the command by name in this part of the docs.

blp:
It's probably fine in many individual cases to refer to the command generically.

We might want to invent a mechanism for cases where we do want to do substitutions, but we should be able to get along without it for some time.

blp:
Grim-lock: Would you mind updating this to resolve the conflicts that have arisen since it was posted?

russellb:
and please leave a comment on here if you post an updated patch, as github seems to only email me on comments, not when patches are updated :-/

blp:
Grim-lock, are you abandoning this?

Grim-lock:
Sorry for delay, I will resolve the conflicts as quickly as possible.

Grim-lock:
Conflicts has been resolved, pls check, :)

russellb:
Thanks @Grim-lock, I will review this today.

russellb:
Thanks for the patch! I pushed this to master with the following changes:

1) Updated lib/automake.mk so that the build would succeed.
2) Stripped windows line endings from lib/db-ctl-base.xml

PR #94, opened 2015-12-04
Option to enable or disable File Logging for a Running Daemon

Dinesh-Ramakrishnan:
Currently If we want a File Logging feature for a daemon, it needs to be enabled while the daemon is started using --log-file.  We have no option to enable it on the fly while the Daemon is running.  This commit will provide a ovs-appctl command to enable/disable file logging on the fly. 

###Enable File Logging
ovs-appctl -t <target> vlog/enable-file-log [logfilepath]

###Disable File Logging
ovs-appctl -t <target> vlog/disable-file-log


russellb:
Can you try rebasing this patch?  Something isn't quite right with your pull request so it's showing much more than your patch.

blp:
OVS intentionally omits the ability to set the log file at runtime as a security measure.  I was concerned that this would make it possible for anyone who can connect to the unixctl socket to overwrite any file on the file system.

This might be too conservative, but it didn't seem like a useful enough feature to worry about the details.

russellb:
Would this feature be useful at all without the ability to specify a file to use?  That would eliminate the security concern, I think.

blp:
I assumed it was about more than just enabling and disabling file logging, since you can already do that with "ovs-appctl vlog/set".

Dinesh-Ramakrishnan:
@russellb  @blp Thank you for the feedback.  I will close this request and reopen another with proper re-basing.  
The feature I am trying to achieve is as follows. 

Currently we are starting all the daemons without file logging (ie., we dont specify --log-file in the service file).  At later point in time if the user wants to have a log to file (say for a short duration), the only option we have now is to modify the service file and restart the daemon.  Is there any other way to achieve this feature?

Meanwhile we also thought about the following design, 
Enable file logging in the service file and set the file log severity to OFF.  And then on need enable it using vlog/set, but we felt it would be difficult to maintain different severity for different modules in the same daemon. 

Please let me know your suggestion, I will implement accordingly.  
Also is there any way to over come the security issue on unixctl socket.  Can I remove the option of specifying custom log file and always set to the default /var/log/openvswitch/daemon.log? would this be safer?

blp:
I posted a series that adds a vlog/close command:
http://openvswitch.org/pipermail/dev/2015-December/063149.html
http://openvswitch.org/pipermail/dev/2015-December/063150.html
Will you review it?

Dinesh-Ramakrishnan:
@blp I have the following questions on the patch .
1) In the vlog_unixctl_close function the log_file_name was not freed, is this meant to reuse the filename in the reopen command?
2) We have a Scenario where --log-file wont be given in the service file.  In such scenario shall we use the reopen command to enable the file logging to the default log location (/var/log/openvswitch/<daemon>.log)? I think if we have to set the log_file_name to the default location when there is --log-file option is given, 
3) When we are updating the minimal level for logging using update_min_level, wont it affect other Destinations also? I think we should not alter the settings for Console or Syslog.

Please correct me if my assumptions are wrong.

blp:
1) Yes
2) You have to enable a log file when you start the daemon.
3) No, it won't, did you read what update_min_level() does?

blp:
Patches that implement this feature, though not exactly as in this pull request, were applied some time ago.

PR #95, opened 2015-12-11
ovsdb-idl: Add support for column tracking in IDL

shadansari:
Recent IDL change tracking patches allow quick traversal of changed
rows. This patch adds additional support to track changed columns.
It allows an IDL client to efficiently check if a specific column
of a row was updated by IDL.

Signed-off-by: Shad Ansari <shad.ansari@hpe.com>


blp:
Applied to master, thanks!

PR #97, opened 2015-12-15
fix bug for “ovs-ofctl dump-meters br0 -O openflow13”

weizj:
when exec ofputil_put_bands function，the memory will be re allocated for length，so the pointer of reply->length will become invalid. “ovs-ofctl dump-meters br0 -O openflow13” will be reported BAD_LENGTH in the execution.


blp:
Thanks for the report.  I prefer to fix this in a slightly different way, so I sent out a patch for review on the ovs-dev mailing list: http://openvswitch.org/pipermail/dev/2015-December/063466.html
If you'd like to review it, I'd appreciate it.

PR #98, opened 2015-12-21
fix bug :Memory leak

xushengping:


blp:
This seems like a really bad idea because it's freeing data that will be referenced again later.

PR #100, opened 2015-12-22
fix bug :Memory leak

xushengping:


blp:
Thanks for reporting the problem.  I prefer a slightly different fix, which I applied to master: https://github.com/openvswitch/ovs/commit/b0656bbc1e393f5012b858611d274e031760b4bf

PR #102, opened 2015-12-24
fix bug :Release mutex resource in error brance

xushengping:
Signed-off-by: xushengping <shengping.xu@huawei.com>


blp:
Thanks for the bug fix.  I applied this to master, branch-2.5, branch-2.4, and branch-2.3.

PR #104, opened 2015-12-31
ovs-sandbox: assign default value  to ${MAKE}

qhjindev:
When run ovs-sandbox directly from the tutorial directory (the second method
in Tutorial.md), the following result show up:

$ ./ovs-sandbox -b /media/sda6/network/ovs/build
./ovs-sandbox: line 304: install-man: command not found

The reason is that ${MAKE} is not set in the situation.

So change ${MAKE} to ${MAKE-make} to resolve the issue.

Signed-off-by: Qinghua Jin <qhjin_dev@163.com>

### DIFF
--- a/tutorial/ovs-sandbox
+++ b/tutorial/ovs-sandbox
@@ -301,7 +301,7 @@ OVS_SYSCONFDIR=$sandbox; export OVS_SYSCONFDIR
 
 if $built; then
     # Easy access to OVS manpages.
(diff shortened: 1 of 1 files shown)


qhjindev:
I originally sent patch email to dev@openvswitch.org through my email account qhjin_dev@163.com, but not succeed. So now i use github to commit.

russellb:
Thanks for the patch!  I pushed this to master.

PR #105, opened 2016-01-07
datapath: Fix datapath build on Centos 6.5 (2.6.31-431) kernels

sabyasg:

PR #106, opened 2016-01-07
datapath: Fix datapath build on Centos 6.5 (2.6.31-431) kernels

sabyasg:


blp:
This appears to duplicate pull request #105, so I'm closing it.

PR #109, opened 2016-02-23
Add --enable-jemalloc argument to build

estebarbhpe:
During our tests with OVSDB we found out that memory allocation is
used intensively by the server, so we thought that using a different
memory allocator could increase the performance. We tried jemalloc
and the performance gain was between 20% and 40%.

This patch would allow anyone to enable jemalloc on their systems
in case you want to try out this memory allocator.

The patch adds the --enable-jemalloc flag to link ovsdb-server with jemalloc
instead of the default memory allocator. This can improve the OVSDB
Server performance under certains loads.

Signed-off-by: Esteban Rodriguez Betancourt <estebarb@hpe.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ addons:
       - gcc-multilib
       - libssl-dev
       - llvm-dev
+      - libjemalloc1
+      - libjemalloc-dev
 
 before_install: ./.travis/prepare.sh
 
@@ -33,6 +35,7 @@ env:
   - KERNEL=3.4.110
   - KERNEL=3.2.76
   - KERNEL=2.6.32.70
+  - OPTS="--enable-jemalloc"
 
 script: ./.travis/build.sh $OPTS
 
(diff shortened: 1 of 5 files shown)


russellb:
Have you tried just using LD_PRELOAD to enable jemalloc?

You show some performance results, which is fantastic.  An improvement would be make those tests easily repeatable by others.  There are many reasons others may want to run the same tests (see how results compare on different systems, test with ovsdb-server changes, test with other allocators ...)

estebarbhpe:
We used LD_PRELOAD, but at the end preferred this solution, because we want it to crash at compile time if jemalloc isn't present.
The benchmark tool used is here: https://github.com/estebarbhpe/ovs/tree/feature/benchmark . That tool is still WIP. Our intent is to publish the benchmark tool too, but it still need changes.

estebarbhpe:
Closing this pull request and creating https://github.com/openvswitch/ovs/pull/112 instead.

PR #111, opened 2016-03-08
C IDL Compound Indexes

estebarbhpe:
Adds compound indexes to the C IDL. This improvements allows to do fast lookups over columns other than the UUID, and retrieve rows in order.


blp:
Review is ongoing in threads in ovs-dev.  I'll look forward to v3.

PR #112, opened 2016-03-15
install.md: Suggest jemalloc memory allocator

estebarbhpe:
Change installing documentation to suggest to use
jemalloc memory allocator.

This memory allocator showed great performance gains
when used at ovsdb-server and other components.

Signed-off-by: Esteban Rodriguez Betancourt <estebarb@hpe.com>


blp:
Applied.

PR #114, opened 2016-04-07
cksum: Refine schema cksum validation

estebarbhpe:
Calculates the cksum removing the cksum line using a more
strict regex than the used previously.
It fixes a problem when calculating the cksum of a schema that
has fields with the substring cksum (e.g.: a checksum column),
lines that the previous cksum calculation incorrectly removes
before running cksum.


blp:
Apologies for the delay in review--it was far too long, of course.

If this is still relevant, then please resubmit it as a single commit.  There should be no need for 8 commits to add this small feature.

blp:
Oops, good to know, thanks.

PR #115, opened 2016-04-18
Fix race condition between networking and openvswitch-switch

adidenko:
Networking upstart service checks if openvsiwtch-switch service
is running (and starts it if it's not running) via
debian/ifupdown.sh hook. But service may be in 'start/pre-start'
state in which it's loading kmod and not yet able to connect to
the database. Which leads to failures in interfaces configuration.

In order to fix this, ifupdown.sh script should wait until
ovs-vsctl is fully functional (able to connectto DB) before trying
to add any records to the database.


adidenko:
Whoops.. my bad, didn't notice you guys are on systemd already and I faced this bug with upstart. So may be not confirmed. Closing this request.

PR #117, opened 2016-04-18
Added support for ofp_packet_out messages in bundles

AndreMantas:
Tested with `make check` and external controller adding packet_out messages to a bundle with different destinations (hosts and controllers). Also tested grouping packet_out with port_mod messages in the same bundle. After committing the bundle, the destinations received the packet_out. 

Uses the same mechanisms and validations of handling a regular packet_out message.

Signed-off-by: Andre Mantas <amantas@lasige.di.fc.ul.pt>


blp:
Jarno reviewed this on the mailing list: https://patchwork.ozlabs.org/patch/639626/

PR #122, opened 2016-05-02
vtep/ovs-vtep: support multiple ovs-vtep processes which using the isolated logical-switch.

nickcooper-zhangtonghao:


blp:
We never merge old branches into master.

PR #124, opened 2016-05-02
Partial Map Updates v2

aymerice:
In the current implementation, every time an element of either a map or set column has to be modified, the entire content of the map/set is sent to the server to be updated. This improvement allows to partially update a map, sending just the modifications (inserts, deletes or updates) to the OVSDB server.

This submission includes a fix to a bug found in a previous submission, in which the IDL crashed when a new row was created and used as reference inside a map in the same transaction.


blp:
Applied to master, thanks!  See http://openvswitch.org/pipermail/dev/2016-May/071231.html for details.

PR #128, opened 2016-05-10
Ovsdb replication patch

cabmario:
The database replication functionality is designed to provide "fail
over" characteristics. There are two participating databases, one of
which is the "active" database and the other is the "stand by" database.
Replication happens exclusively from the active to the stand by
database.


blp:
I posted these patches to ovs-dev so that they can be reviewed in the way that we're most accustomed to.  Stand by for reviews.

The first patch is here: http://openvswitch.org/pipermail/dev/2016-June/073676.html

PR #132, opened 2016-06-09
added toc using https://github.com/ekalinin/github-markdown-toc.go

jacec:
Not a big change, just added table of contents to FAQ as I found it useful.

FYI - it was auto generated using https://github.com/ekalinin/github-markdown-toc.go (which as seen, seems to work) if someone wants to add a TOC to other documentation.


jacec:
I noticed the travis build failed. I took a fork a couple of days ago and most likely inherited a problem master branch. Given my change is FAQ.md only, I still believe this can be merged safely.

blp:
I like the idea of this patch but it seems hard to maintain.  Maybe you could add a Makefile rule to regenerate it automatically.

jacec:
That's possible, but not sure the tool supports a minimal dif when re-run. May need to mark the TOC with start and end tags, delete the lot and re-insert (off the back of a makefile rule - as you suggest). I'll give it some thought, but open to suggestions.

PR #133, opened 2016-06-10
ovsdb: Strong references change cascade fix

estebarbhpe:
Improves the performance of OVSDB avoiding the chain
reaction produced when modifing rows with a strong
reference and the pointed rows have more strong
references.

The approach taken was using the change bitmap to avoid
triggering a change count when the column hasn't changed.

One way to trigger the issue is emulating a simple linked list
with strong references within a table, where each new row
points to the previous.

Without the fix OVSDB creates a ovsdb_txn_row (and a copy
of the row) for each row in the table.
With the fix it only creates two ovsdb_txn_row: the new row and
the directly pointed row.


blp:
Applied to master, thanks!

PR #134, opened 2016-06-27
ovsdb: weak references fix

estebarbhpe:
Weak references performance fix

Prevents the cloning of rows with outgoing or
incoming weak references when those rows aren't
being modified.

It improves the OVSDB Server performance when
many rows with weak references are involved
in a transaction.


blp:
I reviewed this in email: http://openvswitch.org/pipermail/dev/2016-July/074327.html

PR #137, opened 2016-07-03
ovn-nbctl: the improvement to the ovn-nbctl manpage.

nickcooper-zhangtonghao:
If there are multiple logical switches or routers with a duplicate name,
the configuration is slightly different. You should configure the logical
switches or routers using the UUID instead of the name.

Signed-off-by: nickcooper-zhangtonghao <nickcooper-zhangtonghao@opencloud.tech>


blp:
I applied this patch to master.  Thank you!

PR #140, opened 2016-07-04
ovsdb: Fix OVSDB disconnect replication bug

cabmario:
Currently disconnecting from the replicator server means closing the jsonrpc
connection and destroying the monitored table names and blacklisted table
names.

This patch makes a distition between disconnecting from the remote server,
applicable when the replication incurs in an error, and destroying the
remote server info, applicable when ovsdb-server exits gracefully.

Signed-off-by: Mario Cabrera <mario.cabrera@hpe.com>


azhou-nicira:
Pushed with some minor adjustment.

PR #142, opened 2016-07-19
Add sync-from option to ovn-central service

dlevy-ibm:
Adding an option to pass the --sync-from flag to the northbound database
when starting the ovn-central service. This allows database replication
to occur by setting up a parameter in /etc/default/ovn-central.

Signed-off-by: Daniel Levy dlevy@us.ibm.com


jayhawk87:
This needs a better commit message and a Signed-off-by in the commit message (commit -s will produce that for you).  Once you have those, you should send email to dev@openvswitch.org to let folks know about this pull request because most patches are submitted via git format-patch and git send-email.

dlevy-ibm:
Is the CI error legitimate? Can't really make heads or tails of it.

jayhawk87:
you failed in a unit test case - it might not be legitimate, but I don't know how to get a recheck

blp:
My apologies for the unacceptably long review cycle on this patch.

I believe that, in the meantime, support has been added to ovn-ctl for active-backup.

PR #148, opened 2016-08-13
Strip quotes in SYSTEM_VERSION from os-release

mmulsow:
The /etc/os-release file on Ubuntu 14.04 contains:
VERSION_ID="14.04"

Currently this leads to an error starting openvswitch-switch saying:
ovs-vsctl: ""14.04"": quoted string may not include unescaped "

This causes ovs-vswitchd to not get started. With this fix the quotes
will be removed from VERSION_ID and SYSTEM_VERSION will be set to the
correct value of just 14.04.

Signed-off-by: Matt Mulsow <mamulsow@us.ibm.com>

### DIFF
--- a/utilities/ovs-ctl.in
+++ b/utilities/ovs-ctl.in
@@ -523,8 +523,8 @@ set_defaults () {
         SYSTEM_TYPE=`cat $type_file`
         SYSTEM_VERSION=`cat $version_file`
     elif test -e "@sysconfdir@/os-release"; then
-        SYSTEM_TYPE=`awk -F= '/^ID=/{print $2}' @sysconfdir@/os-release`
-        SYSTEM_VERSION=`awk -F= '/^VERSION_ID=/{print $2}' @sysconfdir@/os-release`
+        SYSTEM_TYPE=`. "@sysconfdir@/os-release" && echo "$ID"`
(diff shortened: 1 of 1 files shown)


blp:
How about this version instead? It handles arbitrary shell constructs, not just double-quotes: http://openvswitch.org/pipermail/dev/2016-August/077926.html

mmulsow:
That looks much better. Do you want me to make that change here or abandon this one?

mmulsow:
I've made a very similar change but instead of sourcing /etc/os-release in the current process I call out to a separate shell. That should avoid accidentally changing other variables in this script that happen to conflict with variable names from /etc/os-release.

Also, I'm testing this on my Ubuntu system now. Will post back with results.

mmulsow:
Alright, works well on my Ubuntu 14.04 system. I also tried with bash instead of sh which gives a little bit of coverage for RHEL and most other systems that link /bin/sh to bash (as opposed to Ubuntu using dash).

blp:
I don't think it's necessary to use a separate shell because assignments in a subshell don't propagate to the parent shell.  Try this:
a=b; $(a=c); echo $a
and see that it prints "b", not "c".

mmulsow:
Alright, dropped the 'sh -c'.

blp:
I think that this is now exactly the same as the version I proposed, so I'll take it as a positive review.

blp:
Applied.

PR #152, opened 2016-08-29
Avoid segfault in sFlow when datapath input port for tunnel sample is unknown.

sflow:
The crash is documented here:
http://openvswitch.org/pipermail/discuss/2016-August/022513.html

Signed-off-by: Neil McKee <neil.mckee@inmon.com>


blp:
Applied.

PR #154, opened 2016-09-09
Use debhelper for DEB_UPSTREAM_VERSION

SpamapS:
This variable is set incorrectly if a downstream package builder adds
a + sign or probably other characters. This leads to an unusable
openvswitch-datapath-source package and unbuildable dkms modules.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,11 @@
+openvswitch (2.7.90-2) UNRELEASED; urgency=medium
+
+  * Use debhelper for DEB_VERSION_UPSTREAM
+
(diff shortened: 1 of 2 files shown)


blp:
My apologies for the unacceptably long review time on this pull request.

I believe that this problem has been fixed in the meantime with commit 455f38e009203795c1a27673969a186002564dde.  However, your version is better.  I'd like to use it instead.

Can you provide a Signed-off-by?

Thanks,

Ben.

SpamapS:
Done. :)

blp:
Thanks a lot.  I cherry-picked this to master with commit message updates to reflect that it is more of an improvement than a bug fix at this point.

PR #155, opened 2016-09-18
netdev: Fix netdev->reconfigure_seq memory leak.

hanxueluo:
netdev: Fix netdev->reconfigure_seq memory leak after failed to construct netdev.


blp:
Thanks for the patch.  I applied this to master and branch-2.6.

PR #156, opened 2016-09-21
Remove execute permisson on source files

hanxueluo:


russellb:
Thanks for the patch!  I applied this to master.

blp:
I've applied this to branch-2.6 now.

PR #163, opened 2016-12-06
rconn: Avoid assertion failure if peer fails to send hello.

blp:
If an rconn peer fails to send a hello message, the version number doesn't
get set.  Later, if the peer delays long enough, the rconn attempts to send
an echo request but assert-fails instead because it doesn't know what
version to use.  This fixes the problem.

To reproduce this problem:

    make sandbox
    ovs-vsctl add-br br0
    ovs-vsctl set-controller br0 ptcp:12345
    nc 127.0.0.1 12345

and wait 10 seconds for ovs-vswitchd to die.  (Then exit the sandbox.)

Reported-by: 张东亚 <fortitude.zhang@gmail.com>
Signed-off-by: Ben Pfaff <blp@ovn.org>

### DIFF
--- a/lib/rconn.c
+++ b/lib/rconn.c
@@ -558,19 +558,23 @@ run_ACTIVE(struct rconn *rc)
 {
     if (timed_out(rc)) {
         unsigned int base = MAX(rc->last_activity, rc->state_entered);
-        int version;
-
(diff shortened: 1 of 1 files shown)


blp:
This was committed as commit a203ce69c96e8a38c71113973caf64a22003616f.

PR #166, opened 2017-01-18
Branch 2.6

jerry1001:
hi , We have some questions about the openairinterface(lte) and ovs. . Now we are using RYU try to match tunnel id this field, but can't work . Are the openflow tunnel id and GTPv2 teid the same thing? if  yes what can we solve this problem? thanks


shettyg:
Please use discuss@openvswitch.org for your questions.

PR #170, opened 2017-03-03
id-pool: Fix allocation overflow and the leak of the first id

MarginC:
id_pool_alloc_id() could always allocate (pool->n_ids + 1) ids rather than pool->n_ids ids, it worked well while pool->n_ids = 0.


blp:
Thanks for the patches.  Would you mind resubmitting without duplicative commits and with commit messages properly wordwrapped at about 75 columns?

PR #172, opened 2017-03-20
Fix build with musl libc

kraj:


blp:
Thank you for the patches.

I applied the one for WAIT_ANY to master, although I changed to just use -1 literally instead of conditionally defining WAIT_ANY.

The other patch would break RHEL 6.1: see commit 55bc98d6cb340626eab68fa91eeffafe5e5a2339.  If you can suggest a way to support both, then that would be welcome.

PR #174, opened 2017-03-26
rhel: Fix rpmbuild error on CentOS 6.8

zakinko:
Add support for RHEL6.8 which uses an old version of kernel with some new features backported.


russellb:
If you're using the latest OVS, any reason you're not also using CentOS 7 instead of CentOS 6?

I'm thinking we shouldn't apply this.  OVS currently only intends to support kernels 3.10 and up.  I realize the CentOS 6 kernel has a lot of backports, but I think if you want to use OVS with CentOS 6, it seems like an easier policy to also require a newer kernel.

zakinko:
I have considered updating to CentOS7. 
However since I am satisfied with combination of CentOS6 andOVS2.5, 
can I apply to 2.5 branch if it is difficult to apply to master branch?

russellb:
Sure, feel free to apply the patch locally, but as I mentioned earlier, I don't think we should try to maintain CentOS 6 support upstream.

PR #175, opened 2017-04-17
Branch 2.5

xiaodan1:
man 8 of ovs-ofctl:
''' 
 sctp_src=port/mask:
Bitwise matches on transport ports are rarely useful in isolation, but a group of them can be used  to  reduce  the
              number  of  flows  required to match on a range of transport ports.  For example, suppose that the goal is to match
              TCP source ports 1000 to 1999, inclusive.  One way is to insert 1000 flows, each  of  which  matches  on  a  single
              source port.  Another way is to look at the binary representations of 1000 and 1999, as follows:
              01111101000
              11111001111
              and then to transform those into a series of bitwise matches that accomplish the same results:
              01111101xxx
              0111111xxxx
              10xxxxxxxxx
              110xxxxxxxx
              1110xxxxxxx
              11110xxxxxx
              1111100xxxx
'''

My question is :
  How to transform those into a series of bitwise matches?


blp:
This pull request doesn't make any sense.

PR #183, opened 2017-06-26
tnl-ports.c: Fix bug that ovs-dpdk has no listening tunnel ports when delete one tunnel port.

zhaozhanxu:
Bug is below:

1. When ovs-dpdk has some tunnel ports, I can see the listening ports through the command `ovs-appctl tnl/ports/show -v`.
2. When one tunnel port is deleted, the listening ports are deleted also.
3. This will cause ovs-dpdk can't de-encapsulates packets, then network barrier between two hosts finally.

Signed-off-by: zhaozhanxu <zhaozhanxu@163.com>

### DIFF
--- a/lib/tnl-ports.c
+++ b/lib/tnl-ports.c
@@ -51,6 +51,7 @@ static struct ovs_list addr_list;
 
 struct tnl_port {
     odp_port_t port;
+    struct ovs_refcount ref_cnt;
     ovs_be16 tp_port;
     uint8_t nw_proto;
     char dev_name[IFNAMSIZ];
@@ -195,7 +196,8 @@ tnl_port_map_insert(odp_port_t port, ovs_be16 tp_port,
     ovs_mutex_lock(&mutex);
     LIST_FOR_EACH(p, node, &port_list) {
         if (tp_port == p->tp_port && p->nw_proto == nw_proto) {
-             goto out;
+            ovs_refcount_ref(&p->ref_cnt);
+            goto out;
         }
     }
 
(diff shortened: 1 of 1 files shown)


blp:
Thank you for the bug fix.

Can you give an example of how the bug would be triggered?  I'm trying to understand the issue.

Thanks,

Ben.

zhaozhanxu:
This bug is triggered in a cluster, the main reason is that ovs-dpdk can not decapsulate packets, because there is no listened 6081 port.

1. We need three host(host1, host2 and host3). Each host connect the other two hosts through geneve ports, that is, each ovs-dpdk have two geneve ports.
2. Configure network makes three vms(one vm per host) can communicate with each other normally. This way, we can see there are some listened 6081 ports on the three hosts through the command `tnl/ports/show -v`, that is, geneve packets can be properly decapsulated. Now the environment is regular.
3. Delete any geneve port of host3 , then vm1(on host1) and vm2(on host2) can not communication. This way, we can see there is no listened 6081 port on host1 and host2, so that the two hosts can not properly decapsulate geneve packets.

Sorry that my English is poor, if you still can't understand, I will try to draw a picture to explain. ^_^

blp:
I understand now.

I updated the commit message and applied this to master.  Let me know if you'd like it applied to older branches as well.

zhaozhanxu:
Thanks.

I Think it can be applied to branch-2.7 and branch-2.6.

blp:
> I Think it can be applied to branch-2.7 and branch-2.6.
OK, I applied it to those branches too.

PR #187, opened 2017-06-27
pkt reassemble: fix kernel panic for ovs reassemble

wangzk320:
Ovs and kernel stack would add frag_queue to same netns_frags list.
As result, ovs and kernel may access the fraq_queue without correct
lock. Also the struct ipq may be different on kernel(older than 4.3),
which leads to invalid pointer access.

The fix creates specific netns_frags for ovs.

Signed-off-by: wangzhike <wangzhike@jd.com>


blp:
I posted this to ovs-dev to make sure that it gets noticed.

joestringer:
Thanks again for investigating this and sorry it took so long to test it out. Unfortunately this code is very prone to affecting rare race conditions, so it is a difficult thing to get confidence about the potential impact in conjunction with different kernel releases.

I applied this:
https://mail.openvswitch.org/pipermail/ovs-dev/2017-July/336087.html

PR #188, opened 2017-06-30
Python2 and Python3 compatibility

jwessel:
The meta-virtualization project would like to be able to use python2 or or only python3 to build and use the openvswitch. 

This set of patches adds compatibility between python2 and python3.  The output of all the generated files was tested before the patches were applied with python2 and after applying the patches with python2 and python3.   Everything matches correctly.  Some additional patches were needed to make the output of the files deterministic.  When the iterators are used on the dict types the state of the heap and timing play a role where the order is not always the same.   For example running the make clean ; make -j20 in a loop will sometimes generate a different lib/ofp-actions.inc2 where the all_raw_instances[] has the static populated struct in a different order. 

If there is interest in making the order of the generated files deterministic it can be done with additional patches that add a sort applied to the dict iteration.  These patches were used to prove the generated files were in fact the same.

### DIFF
--- a/build-aux/check-structs
+++ b/build-aux/check-structs
@@ -212,7 +212,7 @@ def checkStructs():
 
     if '--help' in sys.argv:
         argv0 = os.path.basename(sys.argv[0])
-        print '''\
+        print('''\
 %(argv0)s, for checking struct and struct member alignment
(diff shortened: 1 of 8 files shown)


blp:
Thanks a lot.  I applied all of these patches to the OVS master branch.

PR #190, opened 2017-07-12
lacp: enable bond slave immediately after lacp attach

hanxueluo:
There is a long interval (5~20 seconds) between lacp slave attach
and bond slave enable. During the interval, ovs drop all received
packets from that slave because bond_check_admissibility() check
fails. The root cause is that connectivity_seq is not changed
after lacp update and lacp status is not populated into port->may_enable
by port_run() immediately.

### DIFF
--- a/lib/lacp.c
+++ b/lib/lacp.c
@@ -536,6 +536,7 @@ lacp_run(struct lacp *lacp, lacp_send_pdu *send_pdu) OVS_EXCLUDED(mutex)
 
     if (lacp->update) {
         lacp_update_attached(lacp);
+        seq_change(connectivity_seq_get());
     }
 
     HMAP_FOR_EACH (slave, node, &lacp->slaves) {


mtarsel:
I can confirm this behaviour on ppc64le. @hanxueluo curious which arch you are using, x86?

mtarsel:
I have applied this patch and ran a few tests however I am not seeing a huge performance increase here. Any ideas what would be a more appropriate time? I think the current time range is too much but what should we be aiming for here?

hanxueluo:
Hi, @mtarsel I'm using x86. I'm not sure what your test result is.

Maybe you need to comment out route_table_run() to do the test.  
In some environment, route_table_run() destruct and construct netdev randomly. Its calling netdev_change_seq_changed() frequently makes  it hard to reproduce the 5~20s interval.

mtarsel:
@hanxueluo - thank you. Ill try commenting out route_table_run(). I was just asking arch for documentation purposes although shouldn't matter in this case.
We were doing tests with ovs bonding using network name spaces and the delay we were seeing was due to the ARP table not updating with the MAC address for the newly active slave. The 2 name spaces had the same IP but different MAC so ARP entry needs to be updated - there is no higher-level redundancy management service in this setup. There was a RARP packet sent to the newly active slave interface but the ARP exchange did not happen for some time (the delay I saw being closer to 45 seconds).

 I was curious why RARP was still used in ovs bonding and why there was a not a gratuitous ARP sent out.

blp:
Thanks a lot for the patch.  I applied this to master and backported as far as 2.4.

PR #198, opened 2017-08-21
Fix: vhost user port is displayed as LINK DOWN after ovs-vswitchd reb…

wangzk320:
…oots


wangzk320:
The traffic is OK, but the status is shown as LINK DOWN.

PR #210, opened 2017-10-19
fix 3 bugs about igmp、mirror、xlate

hanxueluo:


blp:
I am sorry that I missed these patches until now.  Some of them look like they fix important bugs.  I have now sent them to the ovs-dev mailing list for discussion.

PR #212, opened 2017-11-06
Portability fix. Change #include order

borjam:
This file doesn't compile on FreeBSD nor OS X because icmp6.h needs definitions from sys/types.h.

On Linux icmp6.h includes sys/types.h itself but it's not a good idea to assume that everyone does. Besides, I think it's much more clear to put it before the netinet stuff so that more "fundamental" definitions are included first.

### DIFF
--- a/lib/conntrack.c
+++ b/lib/conntrack.c
@@ -17,10 +17,10 @@
 #include <config.h>
 #include <ctype.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


blp:
Ugh, another place that FreeBSD makes builds gratuitously fail.

Thanks, I applied this to master.

PR #213, opened 2017-11-21
tunnel: Reset tunnel flags in tnl_port_send().

gasmith-nutanix:
Most tunnel flags (like FLOW_TNL_F_KEY and FLOW_TNL_F_CSUM) should not be
inherited from ingress tunnel to egress tunnel. The setting for these flags
should be derived from the egress tunnel configuration.

Currently, any flag set by the ingress tunnel is incorrectly inheirted by the
egress tunnel. This regression was introduced along with the OAM flag, which is
presumably intended to be inherited.

CC: Jesse Gross <jesse@nicira.com>
Fixes: b666962be3b2 ("tunneling: Allow matching and setting tunnel 'OAM' flag.")
Signed-off-by: Gregory Smith <gasmith@nutanix.com>

### DIFF
--- a/ofproto/tunnel.c
+++ b/ofproto/tunnel.c
@@ -462,7 +462,8 @@ tnl_port_send(const struct ofport_dpif *ofport, struct flow *flow,
         }
     }
 
-    flow->tunnel.flags |= (cfg->dont_fragment ? FLOW_TNL_F_DONT_FRAGMENT : 0)
+    flow->tunnel.flags = (flow->tunnel.flags & FLOW_TNL_F_OAM)
+        | (cfg->dont_fragment ? FLOW_TNL_F_DONT_FRAGMENT : 0)
         | (cfg->csum ? FLOW_TNL_F_CSUM : 0)
         | (cfg->out_key_present ? FLOW_TNL_F_KEY : 0);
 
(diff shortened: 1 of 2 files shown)


blp:
Well, now I feel like an idiot.  We recently re-discovered the same problem and fixed it in the folllowing commit: 

commit 81f9ef5a3dd7da6f93ab64980b3d5eaee5de7058
Author: William Tu <u9012063@gmail.com>
Date:   Wed Jan 10 13:13:11 2018 -0800

    tunnel: fix tunnel flags set/clear.
    
    Existing code only set these tunnel flags (df, csum, and key) when the
    flag is set in the output tunnel port, but did not clear when the flag
    is unset.  The patch fixes it by setting and clearing it accordingly.
    
    Two existing testcases need to fix:
    'tunnel - Geneve option present' has no key set up, so we should match
    'flags(df)' instead of 'flags(df|key)'.  The second case
    'tunnel - concomitant IPv6 and IPv4 tunnels' follows the same pattern.
    One additional test case 'tunnel - Mix Geneve/GRE options' is added.
    
    Signed-off-by: William Tu <u9012063@gmail.com>
    VMWare-BZ: #2019012
    Signed-off-by: Ben Pfaff <blp@ovn.org>

Sorry about the delay and for duplicating work, and thanks for reporting and providing the fix.

gasmith-nutanix:
D'oh! No worries. I'll send patches to the mailing list in the future.

PR #215, opened 2017-12-20
ovn.at generates non-canonical MAC addresses

ryzhyk:
This is a very minor issue, but I guess there is no requirement for ovn-nbctl to accept MAC addresses in non-standard representation?


blp:
Thanks, I applied this to master and branch-2.9.

PR #218, opened 2018-01-31
OpenVSwitch 2.8.1 version, can support kernel-4.14.14 version，or later version support, thank you.

xulongju:
Hi：
OpenVSwitch 2.8.1 version, can support kernel-4.14.14 version，or later version support, thank you.


joestringer:
You can use the openvswitch kernel module that comes with your kernel; it's probably newer than the version in the OVS 2.8 tree anyway.

PR #221, opened 2018-03-20
Update docker run command

frederictobiasc:
Faucet config is now in: /etc/faucet/ and log in: /var/log/
The invocation is updated to comply with [faucet documentation](https://github.com/faucetsdn/faucet/blob/master/docs/docker.rst)

### DIFF
--- a/Documentation/tutorials/faucet.rst
+++ b/Documentation/tutorials/faucet.rst
@@ -147,7 +147,7 @@ between one and the other.
 
 4. Create a container and start Faucet::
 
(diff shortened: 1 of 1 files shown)


blp:
Thanks for the fix.  May I have a Signed-off-by for this contribution?

frederictobiasc:
Hi, sure:
Signed-off-by: Frédéric Tobias Christ <fchrist@live.de>
Should I also add this into the commit message?

blp:
It is not necessary.  I applied this to master and branch-2.9.

PR #222, opened 2018-03-28
update the command for creating the faucet container

korimas:
The specified directory is wrong.


blp:
Thanks for the patch.  May I have a Signed-off-by for this patch?  See http://docs.openvswitch.org/en/latest/internals/contributing/submitting-patches/

PR #223, opened 2018-04-09
rhel: user/group openvswitch does not exist

apevec:
Default ownership[1] for config files is failing on an empty system:
  Running scriptlet: openvswitch-2.9.0-3.fc28.x86_64
warning: user openvswitch does not exist - using root
warning: group openvswitch does not exist - using root
...

Required user/group need to be created in %pre as documented in
Fedora guideline[2]

[1] https://github.com/openvswitch/ovs/commit/951d79e638ecdb3b1dcd19df1adb2ff91fe61af8

[2] https://fedoraproject.org/wiki/Packaging:UsersAndGroups#Dynamic_allocation

Signed-off-by: Alan Pevec <alan.pevec@redhat.com>

### DIFF
--- a/rhel/openvswitch-fedora.spec.in
+++ b/rhel/openvswitch-fedora.spec.in
@@ -92,6 +92,7 @@ Requires: openssl hostname iproute module-init-tools
 #Upstream kernel commit 4f647e0a3c37b8d5086214128614a136064110c3
 #Requires: kernel >= 3.15.0-0
 
+Requires(pre): shadow-utils
 Requires(post): /usr/bin/getent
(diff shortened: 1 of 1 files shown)


orgcandman:
Thanks, Alan!  I'll submit this to the ovs-dev list.

krsacme:
It is missing the creation of "hugetlbfs" group, which is required for dpdk, will there be a follow up?

apevec:
Please review
[PATCH v3] rhel: user/group openvswitch does not exist
https://mail.openvswitch.org/pipermail/ovs-dev/2018-April/346315.html

PR #224, opened 2018-04-12
edit Open vSwitch license info so that GitHub recognizes it

eirinikos:
Hello! 👋

(CCing @dankohn who has requested this work so that the license will appear correctly in https://landscape.cncf.io/selected=open-v-switch)

GitHub uses a library called Licensee to identify a project's license
type. It shows this information in the status bar and via the API if it
can unambiguously identify the license.

This commit creates a LICENSE file that stores the full text of the
Apache license. It also removes the COPYING file and transfers its
contents to a new "License" section in the README.

Collectively, these changes allow Licensee to successfully identify the
license type of Open vSwitch's codebase as Apache.


blp:
Thanks.  I made some updates and submitted the updated version to our developer mailing list for review (which is the more common way that we handle submissions for OVS).  You should be CCed so that you continue to receieve updates.

eirinikos:
Thanks @blp. I'll follow the thread on your mailing list.

PR #230, opened 2018-05-06
ovn/docker-overlay: stretch Logical_Switch across multiple clusters

greenpau:
By default this plugin creates a logical switch on per key-value
store basis. The plugin derives the logical switch name from the
NetworkID, which is cluster specific. When this option is enabled,
the logical switch name is derived from IP subnet and the subnet's
context. This way a logical switch may stretch across multiple
clusters managed by different key-value stores. The context
provides for overlapping address space.

Signed-off-by: Paul Greenberg <greenpau@outlook.com>

### DIFF
--- a/Documentation/howto/docker.rst
+++ b/Documentation/howto/docker.rst
@@ -42,13 +42,30 @@ or later.
 Setup
 -----
 
-For multi-host networking with OVN and Docker, Docker has to be started with a
-distributed key-value store. For example, if you decide to use consul as your
-distributed key-value store and your host IP address is ``$HOST_IP``, start
-your Docker daemon with::
+OVN integrates with Docker using the specification for Docker LibNetwork
+remote plugin. See https://github.com/docker/libnetwork/blob/master/docs/remote.md.
+
+OVN implements two plugins:
+
+- "overlay" mode: ``ovn-docker-overlay-driver``
+- "underlay" mode: ``ovn-docker-underlay-driver``
+
+Typically, for multi-host networking with OVN and Docker, Docker has to be
+started with a distributed key-value store. For example, if you decide to
+use consul as your distributed key-value store and your host IP address
+is ``$HOST_IP``, start your Docker daemon with::
 
     $ docker daemon --cluster-store=consul://127.0.0.1:8500 \
         --cluster-advertise=$HOST_IP:0
+
+However, the key-value store is unnecessary if the plugins, ``NetworkDriver``
+and ``IpamDriver``, used by Docker LibNetwork driver are ``Local`` in
+``Scope``. For example, the reason that the default ``Scope`` in ``overlay``
+mode is set to ``Global`` is IPAM function is provided by Docker native IPAM
+plugin. The native IPAM plugin requires a key-value store to ensure consistency
+of IP address spaces. If the IPAM driver used implements ``Local`` scope, then
+there is no need for key-value store and it is safe to set the plugin's scope
+to ``Local`` via ``--docker-driver-scope local`` command-line argument.
 
 OVN provides network virtualization to containers. OVN's integration with
 Docker currently works in two modes - the "underlay" mode or the "overlay"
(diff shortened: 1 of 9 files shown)


greenpau:
Hi @shettyg , I noticed you wrote the initial version of the plugin. Could you please review this pull request? I would appreciate your feedback and insights.

shettyg:
>Hi @shettyg , I noticed you wrote the initial version of the plugin. Could you please review this pull request? I would appreciate your feedback and insights.

I will take a look. Lots of new code, so it will take a few days.

greenpau:
> I will take a look. Lots of new code, so it will take a few days.

@shettyg , thank you! :-) There are still some additional work there, e.g. key-value store cleanups and synchronization of ports that were removed without Libnetwork calls, e.g. power off, or `kill -9 <docker-PID>`. I just want to get a head start, since there is a lot of code 😄

greenpau:
@shettyg , I also tried leaving the original plugin flow intact. The majority of my additions are optional.

blp:
Doesn't build:

> ../Documentation/howto/docker.rst:327:
>     $ echo -n "default-10.10.10.0/24" | sha256sum
> See above for uses for "echo -n", which is non-POSIX
> and does not work with all shells.  Use "printf" instead.

shettyg:
@greenpau 

Sorry about the delay. I forgot about this (Not having a good TODO list is to blame). Overall, I see that you have thought long and hard to make driver "local" and changing it to fit your use case.  But a change which modifies OVN flow table out of band is something that cannot be maintained in the repo. The OVN flow tables constantly change. The table numbers change with every new feature.

I wrote the docker driver when docker was all the hype. But, after the initial commit, we found it very hard to make the driver flexible to get more data from docker command-line.  Doing a lot of tricks in the background is a little less sustainable in the long run to maintain and so, we decided against pursuing the idea to make the driver use more OVN features.

greenpau:
> But a change which modifies OVN flow table out of band is something that cannot be maintained in the repo. 

@shettyg , even when the features are disabled by default? By default, there are no changes to how the plugin works today.

shettyg:
@greenpau

Yeah. The ideal thing for us is to only handle docker APIs and convert it to OVN APIs. Simple tricks like using IP subnets as network names are okay. Using any backdoor to Docker too is bound to create issues as that may change from one version to another. Using backdoor to OVN is definitely bound to create issues.

greenpau:
> The primary issue is maintainability of code. We don't want wrong code hanging in.

@shettyg , that's fair. Since this PR has multiple features and other additions (e.g. logging, exit handling, cleanups), which of them you consider not a fit?

> The ideal thing for us is to only handle docker APIs and convert it to OVN APIs. 

The practical aspect of it is that once a user powers down a host (or the host crashes) without proper shutdown and it will result in OVS port being abandoned, OVN NB database out-of-sync, etc. Thus, any component that is related to the plugin must be in constant sync, i.e. OVS, Docker, OVN NB. It almost always requires a "shim."

From my perspective, the gap in the "overlay" plugin is that it forces a user to adopt Docker Swarm methodology. The same is true about "underlay" plugin, where you have to adopt OpenStack. It is a move away from modular (which I personally prefer) and move towards "being a part of" Docker Swarm, OpenStack, etc.

For example, why would anyone use OVN IPAM when it is available natively from Docker? On the other hand, if OVN IPAM had a Docker LibNetwork plugin of `IpamDriver` type, then Docker Swarm would not have been necessary and OVN IPAM would have been used instead.

greenpau:
> Doesn't build:

@blp , fixed.

blp:
pychecker errors:

> ../python/ovs/util.py:125:1: E302 expected 2 blank lines, found 1
> ../python/ovs/vlog.py:135:9: E266 too many leading '#' for block comment
> ../python/ovs/vlog.py:136:9: E266 too many leading '#' for block comment
> ../python/ovs/vlog.py:143:48: E703 statement ends with a semicolon

shettyg:
>that's fair. Since this PR has multiple features and other additions (e.g. logging, exit handling, cleanups), which of them you consider not a fit?

For logging, have you looked at vlog.info, vlog.dbg etc? vtep/ovs-vtep.in has some examples.
There are many enhancements we can take. But, it would be nice, if they are split into multiple commits and it is easy pick them and review them. May be cleanup as a commit first and then slowly add new features?

>From my perspective, the gap in the "overlay" plugin is that it forces a user to adopt Docker Swarm methodology. The same is true about "underlay" plugin, where you have to adopt OpenStack. It is a move away from modular (which I personally prefer) and move towards "being a part of" Docker Swarm, OpenStack, etc.

I agree that the plugin as it is now is only useful as a POC for someone to look at then integrate OVN with their inhouse docker infrastructure.

greenpau:
> For logging, have you looked at vlog.info, vlog.dbg etc?

@shettyg , yes. My thinking was that since the service is managed by `systemd` (since added systemd unit file), it would be great to have both `vlog` and stdout. This way journald has all the entries related to the plugin and related services. Appologies for my not knowing/understanding this, are you saying vlog.info can do stdout/stderr?

> But, it would be nice, if they are split into multiple commits and it is easy pick them and review them. May be cleanup as a commit first and then slowly add new features?

@shettyg , I agree, I will split. That's why I want to get feedback on what might be useful for the project. As far as I understand right now, you are not in favor of IP lookups and Docker calls.

> is now is only useful as a POC for someone to look at then integrate OVN with their inhouse docker infrastructure.

@shettyg , 💯 , hence all the work to make it to "alpha" 😄

greenpau:
> pychecker errors ...

@blp, fixed. 

Also, there are other failing tests:

```
2020: simple idl, initially empty, no ops - Python2 - tcp6 FAILED (ovsdb-idl.at:219)
2031: simple idl, initially empty, various ops - Python2 - tcp6 FAILED (ovsdb-idl.at:226)
2042: simple idl, initially populated - Python2 - tcp6 FAILED (ovsdb-idl.at:304)
2053: simple idl, writing via IDL - Python2 - tcp6    FAILED (ovsdb-idl.at:335)
2064: simple idl, writing via IDL with unicode - Python2 - tcp6 FAILED (ovsdb-idl.at:366)
2077: simple idl, handling verification failure - Python2 - tcp6 FAILED (ovsdb-idl.at:412)
2088: simple idl, increment operation - Python2 - tcp6 FAILED (ovsdb-idl.at:443)
2099: simple idl, aborting - Python2 - tcp6           FAILED (ovsdb-idl.at:455)
2110: simple idl, destroy without commit or abort - Python2 - tcp6 FAILED (ovsdb-idl.at:469)
2121: simple idl, conditional, false condition - Python2 - tcp6 FAILED (ovsdb-idl.at:483)
2132: simple idl, conditional, true condition - Python2 - tcp6 FAILED (ovsdb-idl.at:499)
2143: simple idl, conditional, multiple clauses in condition - Python2 - tcp6 FAILED (ovsdb-idl.at:515)
2154: simple idl, conditional, modify as insert due to condition - Python2 - tcp6 FAILED (ovsdb-idl.at:537)
2165: simple idl, conditional, modify as delete due to condition - Python2 - tcp6 FAILED (ovsdb-idl.at:553)
2176: simple idl, conditional, multiple tables - Python2 - tcp6 FAILED (ovsdb-idl.at:580)
2187: self-linking idl, consistent ops - Python2 - tcp6 FAILED (ovsdb-idl.at:619)
2198: self-linking idl, inconsistent ops - Python2 - tcp6 FAILED (ovsdb-idl.at:663)
2209: self-linking idl, sets - Python2 - tcp6         FAILED (ovsdb-idl.at:711)
2220: external-linking idl, consistent ops - Python2 - tcp6 FAILED (ovsdb-idl.at:761)
```

I ran once of them, it fails:

```
## ------------------------------ ##
## openvswitch 2.9.90 test suite. ##
## ------------------------------ ##
2020: simple idl, initially empty, no ops - Python2 - tcp6 skipped (ovsdb-idl.at:219)

## ------------- ##
## Test results. ##
## ------------- ##

0 tests were successful.
1 test was skipped.
```

The error is caused by:

```
OVSDB_CHECK_IDL([simple idl, initially empty, no ops],
  [],
  [],
  [000: empty
001: done
])
```

What is the best way to troubleshoot this? i.e. there is some kind of comparison, how to output what is that baseline being compared to?

blp:
@greenpau Tests generate a testsuite.log with details.  You can read that for full information.

Maybe IPv6 is disabled on your test machine in a way that the tests didn't understand until now.  There was just a fix for that: https://github.com/openvswitch/ovs/commit/f7979b1181132619ea591f2c9e76fa8c37adeb54

greenpau:
> Maybe IPv6 is disabled on your test machine in a way that the tests didn't understand until now. There was just a fix for that: f7979b1

@blp , it is failing in Travis too.

greenpau:
@blp , nevermind, the change was committed today. Will re-sync.

PR #233, opened 2018-05-25
vlog: exit with error if explicitly specified logfile cannot be opened

dcbw:
It seems like if the user wanted a specific logfile but that request
cannot be fulfilled, OVS/OVN shouldn't just continue as if nothing
really happened (besides logging a warning).

Signed-off-by: Dan Williams <dcbw@redhat.com>

@shettyg @rajatchopra @fleitner @putnopvut

### DIFF
--- a/include/openvswitch/vlog.h
+++ b/include/openvswitch/vlog.h
@@ -251,9 +251,14 @@ void vlog_rate_limit(const struct vlog_module *, enum vlog_level,
         case 'v':                               \
             vlog_set_verbosity(optarg);         \
             break;                              \
-        case OPT_LOG_FILE:                      \
-            vlog_set_log_file(optarg);          \
+        case OPT_LOG_FILE: {                    \
+            int err = vlog_set_log_file(optarg);\
+            /* exit if logfile explicitly given \
+             * but cannot be opened. */         \
+            if (err != 0 && optarg)             \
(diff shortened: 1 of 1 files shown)


include/openvswitch/vlog.h
```diff
@@ -251,9 +251,14 @@ void vlog_rate_limit(const struct vlog_module *, enum vlog_level,
...
             vlog_set_verbosity(optarg);         \
             break;                              \
-        case OPT_LOG_FILE:                      \
-            vlog_set_log_file(optarg);          \
+        case OPT_LOG_FILE: {                    \
+            int err = vlog_set_log_file(optarg);\
+            /* exit if logfile explicitly given \
+             * but cannot be opened. */         \
+            if (err != 0 && optarg)             \
+                exit(EXIT_FAILURE);             \
```
jsitnicki: `vlog_set_log_file()` and `vlog_set_syslog_method()` handle errors they cannot recover from by logging a fatal error with `ovs_fatal()`, which in turn terminates the process. Perhaps `vlog_set_log_file()` should be doing the same thing to have consistent fatal error handling across functions from the same family.

blp:
I made some style fixups and applied this to master.  @dcbw Thanks!

@jsitnicki That would be even better.

PR #234, opened 2018-06-05
Removed calls to API deprecated in openssl 1.1

cotequeiroz:
In openssl 1.1, there is no need to initialize the library.  It is automatically done when first used.  This patch allows to compile openvswitch against openssl 1.1.0 with deprecated API disabled.

Signed-off-by: Eneas U de Queiroz <cote2004-github@yahoo.com>

### DIFF
--- a/lib/stream-ssl.c
+++ b/lib/stream-ssl.c
@@ -947,12 +947,14 @@ do_ssl_init(void)
 {
     SSL_METHOD *method;
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined (LIBRESSL_VERSION_NUMBER)
 #ifdef _WIN32
     /* The following call is needed if we "#include <openssl/applink.c>". */
     CRYPTO_malloc_init();
 #endif
     SSL_library_init();
     SSL_load_error_strings();
+#endif
 
     if (!RAND_status()) {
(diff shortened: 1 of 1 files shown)


blp:
Thanks for the patch.

Just above the code that you commented out is a call to CRYPTO_malloc_init() inside #ifdef _WIN32.  Is this now unnecessary too with OpenSSL 1.1+, or is it still needed?

Thanks,

Ben.

cotequeiroz:
The call has changed to OPENSSL_malloc_init.  I suspect that it is not needed anymore, but I can't tell you for sure.  It may be worth investigating.

I'm not very familiar with Windows programming, but reading INSTALL.W32 in openssl-1.0.2, the call is a workaround for compiling applications that don't use multi-threaded DLL.  There is no such warning in 1.1.0.
Both functions (1.0 and 1.1) have a very similar definitions.  The detail that makes me suspect it is not needed--besides the removal of it's mention in the NOTES.WIN file--is that in openssl 1.0.2 `crypto/dh/dhtest.c`, it was called within `#ifdef OPENSSL_SYS_WIN32`, and it is gone in 1.1.0 `test/dhtest.c`.  While there is no clear definition, I will change the commit to include the call only with Openssl  < 1.1.0.

blp:
Thanks, I applied the new patch to master.

PR #236, opened 2018-06-16
Return error code for Rule insert and delete

aravindprasad90:
Currently, rule_insert() and rule_delete() ofproto provider APIs donot
have return values.

There are some possible scenarios where rule insertions and deletions can
fail at runtime even though the static checks during rule_construct() had
passed previously.

Some possible scenarios for failure of rule insertions and deletions:

**) Rule insertions can fail dynamically in Hybrid mode (both Openflow and
Normal switch functioning coexist) where the CAM space could get suddenly
filled up by Normal switch functioning and Openflow gets devoid of
available space.

**) Some deployments could have separate independent layers for HW rule
insertions/deletions and application layer to interact with OVS. HW layer
could face any dynamic issue during rule handling which application could
not have predicted/captured in rule-construction phase.

This patch is the first step to introduce error reporting for rule
insertions/deletions from Client back to OVS.

### DIFF
--- a/ofproto/ofproto-dpif.c
+++ b/ofproto/ofproto-dpif.c
@@ -4444,7 +4444,7 @@ rule_construct(struct rule *rule_)
     return 0;
 }
 
-static void
+static enum ofperr
 rule_insert(struct rule *rule_, struct rule *old_rule_, bool forward_counts)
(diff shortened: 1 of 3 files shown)


blp:
This was already reviewed on the ovs-dev mailing list.

PR #237, opened 2018-06-18
Fix DHCP classless static route for non-classfull masks (i.e. any oth…

RostyslavFridman:
…er than /0, /8, /16, /24, /32)

When trying to determine how many bytes of ip address needs to be included in classless static route option,
we should take into consideration the following.
To get the correct amount of bytes we need to take number of network bits in the mask and divide it by 8.
But if the mask has a remainder when divided, we need to not ignore this and add 1 byte to the to the length of the option.

Signed-off-by: Rostyslav Fridman <rostyslav_fridman@epam.com>

### DIFF
--- a/ovn/lib/actions.c
+++ b/ovn/lib/actions.c
@@ -1667,11 +1667,7 @@ encode_put_dhcpv4_option(const struct ovnact_gen_option *o,
             if (c[i].masked) {
                 plen = (uint8_t) ip_count_cidr_bits(c[i].mask.ipv4);
             }
-            if (plen % 8) {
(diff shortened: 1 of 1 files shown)


blp:
@RostyslavFridman Thank you for the patch.

It looks to me like you could just use DIV_ROUND_UP(plen, 8) in two places, instead of adding extra "if" statements in each place.  Would you mind doing that instead?

Thanks,

Ben.

RostyslavFridman:
@blp Thank you for the suggestion.

Code is updated.

Thanks,
Rostyslav

blp:
Thanks.  I squashed these and pushed them to master and backported as far as branch-2.6.

PR #238, opened 2018-06-20
Update ifupdown.sh

ksanislo:
It seems that line 70 needs to be operating on the $slave variable created in the for loop at :68. 

Bonded interfaces fail to bring up their links with the current version, this will make them work correctly.


shettyg:
It looks right. Can you add a Signed-off-by please to the commit? While at it, can you add your information to AUTHORS file when you send the updated commit.

PR #239, opened 2018-06-20
Correct ifupdown.sh to bring up bonded links properly.

ksanislo:
It seems that line 70 needs to be operating on the $slave variable created in the for loop at :68.

Bonded interfaces fail to bring up their links with the current version, this will makes them work correctly.

Signed-off-by: Ken Sanislo <ken@intherack.com>


blp:
Thanks for the revisions!  I applied this to master and backported it as far as branch-2.7.

PR #243, opened 2018-07-08
Return error code for Rule insert

aravindprasad90:
Currently, rule_insert() ofproto provider API donot have return values.

There are some possible scenarios where rule insertions can fail at runtime even though the static checks during rule_construct() had passed previously.

Some possible scenarios for failure of rule insertions and deletions:

**) Rule insertions can fail dynamically in Hybrid mode (both Openflow and
Normal switch functioning coexist) where the CAM space could get suddenly
filled up by Normal switch functioning and Openflow gets devoid of
available space.

**) Some deployments could have separate independent layers for HW rule
insertions and application layer to interact with OVS. HW layer
could face any dynamic issue during rule handling which application could
not have predicted/captured in rule-construction phase.

Errors in Bundles are not handled in this pull request. Will be handled in next upcoming pull request.

### DIFF
--- a/ofproto/ofproto-dpif.c
+++ b/ofproto/ofproto-dpif.c
@@ -4443,7 +4443,7 @@ rule_construct(struct rule *rule_)
     return 0;
 }
 
-static void
+static enum ofperr
 rule_insert(struct rule *rule_, struct rule *old_rule_, bool forward_counts)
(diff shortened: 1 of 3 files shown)


blp:
Thanks for the patch.  This was submitted on the mailing list so there's no need to have it here also.

PR #244, opened 2018-07-11
Fix crash on FreeBSD

ocochard:
Working on bug https://github.com/openvswitch/ovs-issues/issues/152, I've found wrong mapping of netdev functions on FreeBSD.
While here, remove the invisible character at line 68.

### DIFF
--- a/lib/netdev-bsd.c
+++ b/lib/netdev-bsd.c
@@ -65,7 +65,6 @@
 
 VLOG_DEFINE_THIS_MODULE(netdev_bsd);
 
-
 struct netdev_rxq_bsd {
     struct netdev_rxq up;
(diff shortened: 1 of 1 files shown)


blp:
Would you mind supplying a Signed-off-by?  Thanks.

ocochard:
I've no idea of what a signed-off-by is, how to do it and is it's useful for a one-line patch ?

ocochard:
If it's just a text message, you can add this:
Signed-off-by: Olivier Cochard <olivier@FreeBSD.org>

blp:
@ocochard Yes, that's all I needed.  Thank you, I applied this to master.

PR #247, opened 2018-08-06
rhel: move ovsdb socket to ovsdb-server.socket unit

lkundrak:
When ovsdb-server.service is restarted, there's a time window during which clients aren't able to connect to the socket. To cope with this, systemd is able to manage the sockets and pass it to the services. This patch set makes use of the facility.

### DIFF
--- a/rhel/usr_lib_systemd_system_ovsdb-server.service
+++ b/rhel/usr_lib_systemd_system_ovsdb-server.service
@@ -23,3 +23,4 @@ ExecReload=/usr/share/openvswitch/scripts/ovs-ctl --no-ovs-vswitchd \
            --no-monitor restart $OPTIONS
 RuntimeDirectory=openvswitch
 RuntimeDirectoryMode=0755
+RuntimeDirectoryPreserve=no
--- a/Documentation/ref/ovsdb.7.rst
+++ b/Documentation/ref/ovsdb.7.rst
@@ -423,6 +423,11 @@ punix:<file>
     On Windows, listens on a local named pipe, creating a named pipe
     <file> to mimic the behavior of a Unix domain socket.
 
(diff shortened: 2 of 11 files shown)


blp:
Thank you for working to make OVS integrate better into modern systems.

This series makes me vaguely uncomfortable security-wise because it will allow anyone who can modify OVSDB to make ovs-vswitchd accept OpenFlow or OVSDB connections on arbitrary FDs.  I don't have a particularl vulnerability in mind but it seems like a problem that a creative person might be able to exploit.

This makes me think that we should have some means by which "pfd:" streams can only be specified on the command line, not via OVSDB.  (There isn't a perfect mechanism for that at the moment.)

Any thoughts?

rhel/usr_lib_systemd_system_ovsdb-server.service
```diff
@@ -23,3 +23,4 @@ ExecReload=/usr/share/openvswitch/scripts/ovs-ctl --no-ovs-vswitchd \
            --no-monitor restart $OPTIONS
 RuntimeDirectory=openvswitch
 RuntimeDirectoryMode=0755
+RuntimeDirectoryPreserve=no
```
orgcandman: We need to have this either as 'yes', or 'restart' - OVN daemons depend on this directory persisting even when the OVS daemons go away.

drizzt: Since the default is "RuntimeDirectoryPreserve=no", it's probably useless to specify it.
Moreover RuntimeDirectoryPreserve directive is not supported, and so ignored, on old systemd versions (< 235), including the RHEL7 one

horms:
Hi,

Thanks for your contribution.

PR #248, opened 2018-09-12
Documentation: cosmetic fix for example flows

Gaasmann:
Just a cosmetic change on the list of 4 flows at the beginning of the page.
The problem was that with a big enough browser, the first two lines got merged on a single line.

Putting those 4 lines into a block avoid the line merging.
That said, this style is used for code. I let you judge if you find it ok.


Gaasmann:
I modified the commit. I hope it is what expected. :-)
Thanks

blp:
Thanks a lot.  I applied this fix to all affected branches.

PR #249, opened 2018-09-19
ovs-kmod-ctl: source ovs-lib dynamically

javacruft:
Determine installation location of ovs-lib using runtime location
of script, rather than build-time parameters.

### DIFF
--- a/utilities/ovs-kmod-ctl.in
+++ b/utilities/ovs-kmod-ctl.in
@@ -14,7 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


orgcandman:
Thanks for the contribution.  

I originally did think about implementing this in the script.  I somehow figured that a problem with dynamically sourcing this is if the user can trick someone into executing this code and modifying the environment / labels.  This script has an selinux label that allows loading kernel modules.  I need to think about this a bit and get convinced that it won't be a problem.

blp:
This was applied a few days ago.

PR #251, opened 2018-09-27
doc:6 is the TCP protocol number

zhangkaiheb:
17 is the wrong TCP protocol number, changed to 6 in file Documentation/topics/datapath.rst.

Signed-off-by: kai zhang <zhangkaiheb@126.com>


blp:
Thanks, applied to master.

PR #253, opened 2018-10-01
netdev-linux: respect configure result when including net/if_packet.h

armcc:
The configure script already tests whether net/if_packet.h is
available. Respect the test result to fix build issues with musl
libc.


blp:
I don't think that #include is good for anything anyhow.  How about the following instead: https://patchwork.ozlabs.org/patch/978591/

armcc:
Alternative patch looks good.

PR #255, opened 2018-10-02
netdev-linux: use unsigned int for ifi_flags temporary variables

armcc:
ifi_flags in struct netdev_linux is an unsigned int, therefore use
unsigned int for variables which will hold ifi_flags values.


blp:
Thanks! Applied to master.

PR #256, opened 2018-10-03
append Dockerfile filename to build command

zolij:
Faucet latest version (1.8.23) repository doesn't contain Dockerfile (it has multiple Dockerfiles), so docker build failed. I added Dockerfile.faucet as Dockerfile to the command.

### DIFF
--- a/Documentation/tutorials/faucet.rst
+++ b/Documentation/tutorials/faucet.rst
@@ -130,7 +130,7 @@ between one and the other.
 
 2. Build a docker container image::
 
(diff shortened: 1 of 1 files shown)


zolij:
it seems Faucet related pytest fails during image building. Why don't you use already built images from DockerHub?

blp:
I'd be just as happy to suggest prebuilt images. Do you want to submit a patch to do that?

blp:
Thanks!

PR #257, opened 2018-10-03
m4: handle configuring with PYTHON="/usr/bin/env python" etc

armcc:
It may sometimes be desirable to have the python path determined at
run time, ie by setting PYTHON="/usr/bin/env python" etc.

Unfortunately that currently fails as $ovs_cv_python2 and
$ovs_cv_python3 are not always quoted and therefore fail to parse
correctly if PYTHON / PYTHON2 / PYTHON3 are set to values containing
spaces.

### DIFF
--- a/m4/openvswitch.m4
+++ b/m4/openvswitch.m4
@@ -371,7 +371,7 @@ else:
             fi
           done
         done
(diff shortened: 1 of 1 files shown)


blp:
I applied this to master, thanks!

PR #258, opened 2018-10-31
python-c-ext: Fix memory leak in Parser_finish

elapointe-corsa:
The memory returned by json_parser_finish needs to be freed by the caller.

Signed-off-by: Eric Lapointe <elapointe@corsa.com>
---
This fix can easily be tested using the following script in the python shell and looking at the memory usage with 'top':
```
import ovs.json
input = '{"key": "value"}'
while True:
    p = ovs.json.Parser()
    input = input[p.feed(input):]
    json = p.finish()
```


blp:
Thanks, I applied this to master and backported it as far as branch-2.6.

PR #259, opened 2018-11-06
python-ovsdb: Changes to fetch list type values from ovsdb

black0027:
There is a problem when loading ovs ports list with python-ovsdb lib. "Datum.to_python()" returns None.
For example,

\# ovs-vsctl get <br_name> port
[108beebd-3089-425a-a55d-aedaeddb8107, 92cb0e47-8c7d-4dfc-8480-410eea3411a3, a0cd6f8d-5cbc-433f-b52a-f7671ecba71b, ad9afb13-8246-4a06-bf04-0fca0093012c]

Signed-off-by: Nobody Void <black0027@hotmail.com>

### DIFF
--- a/python/ovs/db/data.py
+++ b/python/ovs/db/data.py
@@ -493,6 +493,11 @@ class Datum(object):
                 if dk is not None and dv is not None:
                     value[dk] = dv
             return value
+        elif self.type.is_list():
+            values = list()
+            for k in self.values.keys():
+                values.append(k)
+            return values
         else:
             s = set()
             for k in self.values:
--- a/python/ovs/db/types.py
+++ b/python/ovs/db/types.py
@@ -500,6 +500,9 @@ class Type(object):
     def is_map(self):
         return self.value is not None
 
+    def is_list(self):
+        return self.key is not None and self.value is None
+
     def is_smap(self):
         return (self.is_map()
                 and self.key.type == StringType


blp:
Can you help me better understand the problem?  The "for example" doesn't actually use the Python OVSDB library.

it doesn't look appropriate to me to use to_string() in the implementation.  What if the list isn't a list of strings?

black0027:
Yes, The "for example" doesn't actually use the Python OVSDB library. 
What I meaned was when trying to use python-ovsdb lib to get the ports list of ovs like as the result of executing '**ovs-vsctl get <br_name> port**', However, in the existing implementation, list type returns None.

And the datum class implements the to_string() method, but if it is not a string I think that can consider changes just like below so that user application can deal object detail.

- values.append(k.to_string())
+ values.append(k)

blp:
OK, please update the patch.

blp:
Please squash the changes in a single patch.

black0027:
Sorry for late. Making a single patch has done.

blp:
I'm sorry it's taken so long to take a look at this.

I would really appreciate it if you would add a test (that fails before the patch and succeeds afterward) or at least a complete example.

blp:
I'm sorry it's taken so long to take a look at this.

I would really appreciate it if you would add a test (that fails before the patch and succeeds afterward) or at least a complete example.

blp:
Actually, this patch makes many of the unit tests fail, so I suspect that it is simply wrong.

PR #262, opened 2018-11-28
ovs-ofctl: update a man page about group syntax

nmi:
`watch_port` and `watch_group` parameters can also be used for groups whose type is `select`, so I updated an ovs-ofctl man page.

Signed-off-by: Nobuhiro MIKI <nob@bobuhiro11.net>


blp:
Thanks, applied to master.

PR #266, opened 2018-12-25
ovn: Fix incorrect comparison of the NB and SB band actions

maksqwe:


blp:
Thanks a lot.  I applied this patch to master and branch-2.10.

PR #268, opened 2019-01-04
Cherrypick the datapath compat fix on branch 2.6 to support 3.10 kernel

liucimin:
Cherrypick the datapath compat fix on branch 2.6 to support 3.10 kernel


blp:
We don't backport features into old branches. You can use the kernel module from a newer release instead.

PR #269, opened 2019-01-17
Fix OpenFlow v1.3.4 Conf test failures: 430.500, 430.510

psiyengar:
This commit adds additional verification to nx_pull_header__()
in lib/nx-match.c to distinguish between bad match and bad action
header conditions and return the appropriate error type/code.

Signed-off-by: Prashanth Iyengar <prashanth_iyengar@alliedtelesis.com>
Reviewed-by: Tony van der Peet <tony.vanderpeet@alliedtelesis.co.nz>
Reviewed-by: Rahul Gupta <Rahul_Gupta@alliedtelesis.com>

### DIFF
--- a/lib/meta-flow.c
+++ b/lib/meta-flow.c
@@ -3503,7 +3503,7 @@ mf_vl_mff_nx_pull_entry(struct ofpbuf *b, const struct vl_mff_map *vl_mff_map,
                         const struct mf_field **field, union mf_value *value,
                         union mf_value *mask, uint64_t *tlv_bitmap)
 {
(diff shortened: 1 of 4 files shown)


blp:
Thank you for the patch!  It is good to make OVS more conformant.

It looks like the new function nxm_validate_action_field_header() is only used within nx-match.c.  I'd prefer to have it be static in that case.

nxm_validate_action_field_header() shares most of its code with nxm_field_by_header().  I'd prefer to see both of them written in terms of a common helper function to reduce duplication.

Thanks again!

blp:
Thanks, I squashed these together and applied them to master.

PR #271, opened 2019-02-12
Use pkg-config for DPDK library

geaaru:
With meson integration and in the next releases in the DPDK library
libdpdk.so and in will no more available. Libraries to link can be
retrieved with pkg-config --libs libdpdk command.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -214,6 +214,12 @@ AC_DEFUN([OVS_CHECK_DPDK], [
                               [Specify the DPDK build directory])],
               [have_dpdk=true])
 
+  # DPDK has two compilation mode: make/gmake or meson (probably main choice in
+  # the next releases). With make/gmake is created an additional library libdpdk.so
+  # that is an hack for simplify users life without use pkg-config.
+  # With meson and with the new releases it is used pkg-config for retrieve
+  # list of libraries to link on DPDK applications. So, if it is used meson
+  # libdpdk.so isn't created.
   AC_MSG_CHECKING([whether dpdk datapath is enabled])
   if test "$have_dpdk" != true || test "$with_dpdk" = no; then
     AC_MSG_RESULT([no])
@@ -240,7 +246,11 @@ AC_DEFUN([OVS_CHECK_DPDK], [
            DPDK_INCLUDE="-I$DPDK_INCLUDE_PATH/dpdk"
         fi
         DPDK_LIB_DIR="$with_dpdk/lib"
-        DPDK_LIB="-ldpdk"
+        DPDK_LIB=$(PKG_CONFIG_PATH=$with_dpdk/build/meson-private/ pkg-config --define-prefix $DPDK_LIB_DIR --libs libdpdk)
+        if test "$?" != "0" ; then
+          # POST: pkg-config fail. I try search for libdpdk.so created with make.
+          DPDK_LIB="-ldpdk"
+        fi
         ;;
     esac
 


istokes:
Support for pkg-config has been upstreamed to OVS master with the followng commit

 5b5aa2d8a54c ("acinclude: Also use LIBS from dpkg pkg-config")

As such I don't think this pull-request is required.

geaaru:
It's only a detail but this: https://github.com/openvswitch/ovs/commit/5b5aa2d8a54c006b6c5239d04b7a751ca5ff5d44#diff-a5522b90382504dc530177f63ced5209R230 will not be available. With Meson libdpdk.so will not be available anymore.

istokes:
Apologies, maybe I've misunderstood, can you elaborate further by what you mean 'libdpdk.so will not be available anymore' ?

Do you mean that meson doesn't create a libdpdk.so? From what I'm aware, 'make' is still the primary tool recommend by DPDK for building, so libdpdk.so should still be available via the usual methods when building DPDK but not through meson.

geaaru:
hi, i spoke with DPDK devs. They said me that meson will be the only way for way for compile DPDK library. Gmake/make will be deprecated.

About libdpdk.so, now this library is a textual tips to avoid user linking of every DPDK libraries (if you open file with vim or another editor you can see something like this: GROUP ( lib1.so lib2.so ).

It only helps the users.

With meson, libdpdk.so is not available anymore because with pkg-config you can retrieve the list of all libraries in a more easy way and so will be dropped.

istokes:
So yes, make will be deprecated, but this is targeted for DPDK 19.11 (at least that was the target provided when I raised the issue at the DPDK Community Release meeting this morning), from what I'm aware 'make' will remain as read only.

I'm also aware that some distros within the community will not be moving to meson quite yet due to dependencies on required packages not be supported in particular packages they provide, so the make method will still have to be used for them in that case.

From the OVS point of view, the next DPDK supported will be 19.11 as OVS sticks to DPDK LTS releases so it's a little while out yet but your right, it's something that will have to be considered when moving to 19.11.

geaaru:
But I think that this depends on distro. If you compile dpdk with meson also on 18.11 you have the same problems.

Probably, my patch could be more efficient if try to handle both worlds (meson and make).

For example, I created ebuild for Gentoo and Sabayon distro with already meson.

istokes:
Sure, but see below taken from the [18.11 meson build notes in DPDK](https://github.com/DPDK/dpdk/blob/0da7f445df445630c794897347ee360d6fe6348b/doc/build-sdk-meson.txt)

NOTE: Compiling and installing DPDK using ``meson`` and ``ninja``, rather
than using ``make`` (GNU make) is EXPERIMENTAL. Official builds of DPDK
should always be done using ``make``, as described in the ``Getting Started
Guide`` documentation, and at "http://core.dpdk.org/doc/quick-start".

As you pointed out, distros may choose to build with meson but with the above caveat considered. Using the existing make functionality will still work for 18.11 (and is required for distros that do not support meson).

Note that the original pull request for your patch caused all [OVS DPDK builds to fail](https://travis-ci.org/openvswitch/ovs/builds/492231747) in the OVS travis job, this would have to be fixed in your patch. 

If you have changes required for the gentoo build to support both make and meson I'd be happy to review them when you submit them, but it would have to provide a  clean build for OVS travis also.

geaaru:
> NOTE: Compiling and installing DPDK using `meson` and `ninja`, rather
> than using `make` (GNU make) is EXPERIMENTAL. Official builds of DPDK
> should always be done using `make`, as described in the `Getting Started Guide` documentation, and at "http://core.dpdk.org/doc/quick-start".
> 

:) This is the reason because I asked to DPDK devs team to change that notes. I will push again for this.

Ok, yeah, I try to prepare the fix in the next days. I will appreciate your help to fix all Travis task.

geaaru:
As additional information for the thread, here the commit related to updating of meson as default technology:

https://github.com/DPDK/dpdk/commit/7c383daae0dd446468675676a8386360b2c81a67

Drop of the Make is yet under discussion.

geaaru:
@istokes let me know WDYT of my changes. Now, it seems all good by Travis. Thanks.

istokes:
Thanks for the update, there's ongoing work on the mailing list around this from other distros over the last week, I'm not sure they;re aware of this, can I ask you to flag the pull request on the mailing list (you can email myself but just cc ovs-dev@openvswitch.org), I dont see an issue but I'd like to ensure others are ok with the change also, if its ok we can add it for the 2.11 release then as well as master.

geaaru:
done ;)

geaaru:
As described on ml a workaround for this is to set PKG_CONFIG_PATH with a custom path of DPDK library installation.

PR #275, opened 2019-02-18
ofproto-dpif-xlate: Learn source MAC for output action

antonkurbatov:
This patch allows to learn a source MAC address for 'output' action.
Which in turn will allow to reduce packets flooding in some cases, e.g.
when the rule for an opposite flow is the 'NORMAL'.

Signed-off-by: Anton Kurbatov <bxfree@gmail.com>

### DIFF
--- a/ofproto/ofproto-dpif-xlate.c
+++ b/ofproto/ofproto-dpif-xlate.c
@@ -566,7 +566,8 @@ struct xlate_bond_recirc {
 
 static void compose_output_action(struct xlate_ctx *, ofp_port_t ofp_port,
                                   const struct xlate_bond_recirc *xr,
-                                  bool is_last_action, bool truncate);
+                                  bool is_last_action, bool truncate,
+                                  bool mac_learn);
 
 static struct xbridge *xbridge_lookup(struct xlate_cfg *,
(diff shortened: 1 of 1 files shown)


blp:
This is clever!  I wonder whether it could have undesired side effects.  For example, suppose that some flows output to multiple ports, perhaps flooding packets "by hand" instead of through outputting to OFPP_FLOOD.  In that case, I think that this new logic will learn the final output port as the source, even though flooding should not do that.

I've occasionally thought about reimplementing OFPP_NORMAL.  It could be done in terms of finer-grained specialized actions, or perhaps in terms of carefully written LEARN actions.  Or it could even just expose the MAC learning table through some OpenFlow table.  In the latter case, LEARN actions  and all the usual OpenFlow flow table manipulation methods could be used to view and adjust the MAC learning table.  That might be a bigger project than you intended to take on, though.

antonkurbatov:
Ben, thanks a lot for your comment.
When I was preparing the patch I just wanted to eliminate packets flooding from NORMAL action in case of opposite flow is OUTPUT:PORT.
As you mentioned above, looks like this patch doesn't cover other cases and may break current flow behaviour.
Going to decline the PR.

PR #278, opened 2019-03-26
test: Fix fragment-related tests that fail due to small-sized packets

yifsun:
These fragment-related tests are failing on later kernels (4.19.x)
because kernel quietly drops any packet fragment that is not the last
but has a size smaller than IPV6_MIN_MTU. This patch fixes
them by increasing their sizes to IPV6_MIN_MTU.

Signed-off-by: Yifeng Sun <pkusunyifeng@gmail.com>


yifsun:
This is a v2 patch for https://patchwork.ozlabs.org/patch/1057307/.
It was created after reviewed by Darrell.

tests/system-kmod-macros.at
```diff
@@ -185,3 +185,8 @@ m4_define([OVS_CHECK_KERNEL_EXCL],
     sublevel=$(uname -r | sed -e 's/\./ /g' | awk '{print $ 2}')
     AT_SKIP_IF([ ! ( test $version -lt $1 || ( test $version -eq $1 && test $sublevel -lt $2 ) || test $version -gt $3 || ( test $version -eq $3 && test $sublevel -gt $4 ) ) ])
 ])
+
+m4_define([CHECK_SMALL_V6_FRAG],
+[
+    OVS_CHECK_KERNEL_EXCL(4, 19, 4, 20)
```
darball: Should we not exclude all versions >= 4.19 ?

This could be done using some large value for max version such as 1000, 0 to approximate 'unlimited'
or
using a special 'unlimited' indicator such as 0, 0 and then adding a check for that value in OVS_CHECK_KERNEL_EXCL

I think we should add a comment to CHECK_SMALL_V6_FRAG stating the change in kernel behavior to have
a hardcoded lower limit enforced.

yifsun:
Thanks Darrell, I will fix it and come up with a new version.

Yifeng

yifsun:
Hi Darrell,

I've made fixes according to your review last time. Could you take a look
if you have time? Thanks.

Pull request is at: https://github.com/openvswitch/ovs/pull/278

These tests passed on the following kernels (both check-system-userspace
and check-kmod):
4.20, 4.19, 4.17, 4.14, 4.4, RHEL 3.10

Best,
Yifeng

darball:
Changes are passing all tests. So these are fine.

darball:
Checksum problems were resolved - thanks.

yifsun:
Thanks Darrell.

darball:
Hi Yifeng

Since the checksum issues in earlier versions of this patch were fixed, it is generally ok to merge.

However, note that previous comments requested that some tests keep smaller packets and those
be skipped for the later kernels that use hardcoded checks.
The reason for this request is twofold - show the flexibility of the userspace data path in tests
and delineate the differences in behavior of different kernel version.
Also there are several tests where DPCTL_SET_MIN_FRAG_SIZE() can be removed due to using
larger packets.

Thanks Darrell

blp:
@darball Thanks for the review.  I applied this patch to master.

yifsun:
Thanks Darrell, I will improve it and create another patch.
Yifeng

PR #293, opened 2019-08-07
Using OpenVSwitch with wolfSSL

dgarske:
New configure argument `--with-wolfssl`. For instructions see `README_wolf.md`.


blp:
This should probably come with some kind of rationale. What is WolfSSL and why would one want to use it? I have never heard of it.

The commit message also lacks a Signed-off-by.

dgarske:
Hi @blp,

This was done for a customer who I cannot name. This reason for using wolfSSL is manyfold, but in this case FIPS 140-2 support in the primary reason. Other reasons include modularity (build-time options), reduced vulnerabilities and commercial grade maintenance and support. We have a full openssl compatibility layer that is being used.

Can you expand on what you mean by signed-off-by? I was hoping someone from the OpenVSwitch project could review and comment on the changes.

```
The wolfSSL embedded SSL library is a lightweight, portable, C-language-based SSL/TLS library targeted at IoT, embedded, and RTOS environments primarily because of its size, speed, and feature set. It works seamlessly in desktop, enterprise, and cloud environments as well. wolfSSL supports industry standards up to the current TLS 1.3 and DTLS 1.2, is up to 20 times smaller than OpenSSL, offers a simple API, an OpenSSL compatibility layer, OCSP and CRL support, is backed by the robust wolfCrypt cryptography library, and much more.
The CMVP has issued FIPS 140-2 Certificate #2425 and #3389 for the wolfCrypt Module developed by wolfSSL Inc.
```

Thanks,
David Garske, wolfSSL

blp:
Q: What's a Signed-off-by and how do I provide one?

    A: Free and open source software projects usually require a contributor to
    provide some assurance that they're entitled to contribute the code that
    they provide.  Some projects, for example, do this with a Contributor
    License Agreement (CLA) or a copyright assignment that is signed on paper
    or electronically.

    For this purpose, Open vSwitch has adopted something called the Developer's
    Certificate of Origin (DCO), which is also used by the Linux kernel and
    originated there.  Informally stated, agreeing to the DCO is the
    developer's way of attesting that a particular commit that they are
    contributing is one that they are allowed to contribute.  You should visit
    https://developercertificate.org/ to read the full statement of the DCO,
    which is less than 200 words long.

    To certify compliance with the Developer's Certificate of Origin for a
    particular commit, just add the following line to the end of your commit
    message, properly substituting your name and email address:

        Signed-off-by: Firstname Lastname <email@example.org>

    Git has special support for adding a Signed-off-by line to a commit
    message: when you run "git commit", just add the -s option, as in "git
    commit -s".  If you use the "git citool" GUI for commits, you can add a
    Signed-off-by line to the commit message by pressing Control+S.  Other Git
    user interfaces may provide similar support.

dgarske:
Hi @blp ,

I will look into the DCO and using the `-s` commit option. I don't see any issue with that.

We have a web page (https://www.wolfssl.com/docs/wolfssl-openssl/) that compares wolfSSL and openssl. This might be helpful for describing additional reasons why someone would want to use wolfSSL.

Thanks, David

blp:
On Mon, Aug 19, 2019 at 03:14:46PM -0700, David Garske wrote:
> I will look into the DCO and using the `-s` commit option. I don't see any issue with that.

Thanks.

> We have a web page (https://www.wolfssl.com/docs/wolfssl-openssl/) that compares wolfSSL and openssl. This might be helpful for describing additional reasons why someone would want to use wolfSSL.

I recommend adding that link to the commit message to help developers
understand the commit.

dgarske:
Hi @blp,

I've pushed the changes to add signed off by. Please let me know if there is anything else.

Thanks,
David Garske, wolfSSL

blp:
> Hi @blp,
> 
> I've pushed the changes to add signed off by. Please let me know if there is anything else.

Thanks for the update.

This doesn't integrate very well with the OVS documentation. Installation guides go in Documentation/intro/install, not in the root, and it should be added in Restructured Text format to match the rest of the documentation and to allow it to be put in the table of contents.

It's not clear to me why WolfSSL and Strongswan installation instructions are included. I would presume that each of those libraries comes with its own installation instructions.  

In the end, it looks to me like the only change need to use WolfSSL in OVS is to configure with --with-wolfssl. I guess that could be mentioned in Build Requirements in Documentation/intro/install/general.rst next to libssl, as an alternative to libssl, and maybe other documentation is not really needed at all.

In openvswitch.m4, I see that there are a number of uses of "==" with the "test" utilitiy.  This is nonportable; use = instead.

I would add an item to the NEWS file.

Thanks,

Ben.

dgarske:
Hi @blp,

Thanks for the excellent feedback! Updates pushed, which should resolve your review comments.

David Garske, wolfSSL

blp:
Thanks for resolving my previous comments.

I pulled this into a local branch and looked at it more carefully.

There are no actual code changes here, except for adding a few #include <wolfssl/options.h>. That's the sort of thing that would be easy to support, if necessary, by simply checking at configure time whether that header was needed to #include the openssl headers. I don't see a reason to introduce a separate --with-wolfssl option for that.

I also don't think it's appropriate to encourage users to link against WolfSSL because it will effectively change the license from Apache2 to GPLv2 for those users.

I don't think I can merge this.

PR #299, opened 2019-11-23
bridge: Update stp,rstp statistics every stats-update-interval seconds.

KrishnaKolakaluri:
Moves the field sec_in_state field from status column in Port table
        to statistics column in the same table.This helps in reducing the number
        of update sent to the controller, with the current mechanism
        every-time there is a change in this field a notification is sent to
        the controller. For this field its every second, since it's just
        counting the number of secs its been in that particular state.
        By moving this under port statistics, this can be configured with the key
        "stats-update-interval" in Open_vSwitch table.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -22,6 +22,12 @@ Post-v2.12.0
      * DPDK pdump packet capture support disabled by default. New configure
        option '--enable-dpdk-pdump' to enable it.
      * DPDK pdump support is deprecated and will be removed in next releases.
+   - STP:
+     * New option status-update-interval for stp/rstp status update is added
+       to control the frequency of the updates. The default is 1 second.
+   - RSTP:
+     * The rstp_statistics column in Port table will only be updated every
+       stats-update-interval configured in Open_vSwtich table.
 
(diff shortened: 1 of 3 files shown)


blp:
It looks like this will leave the database inconsistent sometimes. I don't think it's a good idea to update stp_secs_in_state and stp_state at different times, because that makes it impossible for a reader to interpret them.

KrishnaKolakaluri:
> It looks like this will leave the database inconsistent sometimes. I don't think it's a good idea to update stp_secs_in_state and stp_state at different times, because that makes it impossible for a reader to interpret them.

@blp  Can you please suggest a better way to fix it, without causing too many updates to the controller?

blp:
> > It looks like this will leave the database inconsistent sometimes. I don't think it's a good idea to update stp_secs_in_state and stp_state at different times, because that makes it impossible for a reader to interpret them.
> 
> @blp Can you please suggest a better way to fix it, without causing too many updates to the controller?

port_refresh_stp_status() could refrain from updating the column if the only change is a small delta to stp_sec_in_state().

PR #301, opened 2019-12-07
bridge: Update stp_sec_in_state field every 5s and seperate rstp stats.

KrishnaKolakaluri:
"stp_sec_in_state" field in status column of the Port is causing too
	many updates to the controller. As per the suggestion added a 5s interval
	in between the updates of the field.


KrishnaKolakaluri:
@blp  Can you please review this ? I modified according to your comments in PR#299.

PR #302, opened 2019-12-31
bridge: Split the column updates of rstp statistics and status.

KrishnaKolakaluri:


blp:
Thanks. I applied this to master.

KrishnaKolakaluri:
Thank you @blp .

PR #305, opened 2020-01-15
docs: Correct links in mailing-lists.rst

larsks:
Adds `mailto:` to email address links in mailing-lists.rst. The
existing syntax resulted in broken links of the form
`http://docs.openvswitch.org/en/latest/internals/mailing-lists/<address>`,
which would result in a 404 error.


blp:
Thanks. I applied this to master.

PR #307, opened 2020-02-07
Fd persistence

kot-begemot-uk:
This fixes:

1. The current practice of reading FDs without IO pending on them. This is a penalty of 0.3 microseconds per fd. So for example, ovsdb in OVN southdb with 200 nodes has a guaranteed penalty of 60 microseconds on every transaction. That does add up to quite a number in a large installation.

2. The poll loop to do what it is supposed to do according to its comments. It is supposed to be an implementation where the flags (POLLIN, POLLOUT, etc) are additive according to the comments. That is not the case, because if it gets an event which is just POLLIN it will wipe the POLLOUT from its state and vice versa.

3. Adds epoll support for Linux.

I will kill the older pull request with the incomplete linux only/epoll only patchset.

### DIFF
--- a/lib/fatal-signal.c
+++ b/lib/fatal-signal.c
@@ -271,7 +271,7 @@ fatal_signal_wait(void)
 #ifdef _WIN32
     poll_wevent_wait(wevent);
 #else
-    poll_fd_wait(signal_fds[0], POLLIN);
+    poll_fd_wait(signal_fds[0], OVS_POLLIN);
 #endif
 }
 
--- a/lib/fatal-signal.c
+++ b/lib/fatal-signal.c
@@ -96,6 +96,7 @@ fatal_signal_init(void)
         ovs_mutex_init_recursive(&mutex);
 #ifndef _WIN32
         xpipe_nonblocking(signal_fds);
+        poll_fd_register(signal_fds[0], OVS_POLLIN, NULL);
 #else
         wevent = CreateEvent(NULL, TRUE, FALSE, NULL);
         if (!wevent) {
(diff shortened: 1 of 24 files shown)


lib/poll-loop.c
```diff
@@ -320,11 +371,16 @@ poll_block(void)
...
+#ifndef OVS_USE_EPOLL
     struct pollfd *pollfds;
+#endif
+#ifndef OVS_USE_EPOLL
     HANDLE *wevents = NULL;
+#endif
     int elapsed;
     int retval;
     int i;
+    int counter;
```
dcbw: @kot-begemot-uk builds are complaining this variable is unused.

lib/fatal-signal.c
```diff
@@ -236,9 +237,12 @@ void
 fatal_signal_run(void)
 {
     sig_atomic_t sig_nr;
+    char sigbuffer[_POSIX_PIPE_BUF];
```
dcbw: @kot-begemot-uk on the Windows builds, this constant is undefined:

```
fatal-signal.c
lib/fatal-signal.c(240) : error C2065: '_POSIX_PIPE_BUF' : undeclared identifier
lib/fatal-signal.c(240) : error C2057: expected constant expression
lib/fatal-signal.c(240) : error C2466: cannot allocate an array of constant size 0
lib/fatal-signal.c(240) : error C2133: 'sigbuffer' : unknown size
lib/fatal-signal.c(244) : warning C4034: sizeof returns 0
make[2]: *** [lib/fatal-signal.lo] Error 1
make[2]: Leaving directory `/c/openvswitch'
```

kot-begemot-uk:
Superseded by v2.

PR #309, opened 2020-02-20
bond: Fix  command "ovs-appctl bond/disable-slave"  not in effect

shenjiangc:
After disable one slave, it will be enable by the function "bond_link_status_update". 
the bond log like this : 
    2020-02-18T16:21:43.673Z|18945|bond|INFO|interface dpdk0: disabled
    2020-02-18T16:21:43.673Z|18946|bond|INFO|interface dpdk0: link state up
    2020-02-18T16:21:43.673Z|18947|bond|INFO|interface dpdk0: enabled

### DIFF
--- a/ofproto/bond.c
+++ b/ofproto/bond.c
@@ -1744,6 +1744,10 @@ bond_link_status_update(struct bond_slave *slave)
 {
     struct bond *bond = slave->bond;
     bool up;
+    
+    if (!slave->enabled) {
+        return;
+    }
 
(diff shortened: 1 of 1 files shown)


blp:
This patch seems wrong to me. It appears to prevent a slave, once disabled, from ever being enabled automatically. Slaves can get disabled for more reasons than just the "bond/disable-slave" command. Furthermore, the documentation for bond/disable-slave say:

> This setting is not permanent: it persists only until the carrier
> status of \fIslave\fR changes.

PR #314, opened 2020-03-30
raft: Disable RAFT jsonrpc inactivity probe

winsopc:
With the scale test of 640 nodes k8s cluster, raft DB nodes' jsonrpc
session got closed due to the timeout of default 5 seconds probe.
It will cause disturbance of the raft cluster. Since we already have
the heartbeat for RAFT, just disable the probe between the servers
to avoid the unnecessary jsonrpc inactivity probe.

Acked-by: Han Zhou <hzhou@ovn.org>
Signed-off-by: Zhen Wang <zhewang@nvidia.com>


igsilya:
This patch was merged some time ago as commit db5a066c17bdeaa7ecac08870331ae583f5ddfcc.

PR #315, opened 2020-04-08
ipsec: Fix Strongswan configuration syntax

ballock:
Strongswan seems to have .opt files in the source tree with the dotted option syntax. It seems that up until version 5.6, the syntax was also accepted by Strongswan.

However, the .opt files are converted to .conf files during Strongswan build, and the dotted syntax is no longer accepted by Strongswan (tested on 5.8.2).

The effect was that the ovs ipsec monitor fails to start Strongswan, since that complains with:
`/etc/strongswan.d/ovs.conf:4: syntax error, unexpected ., expecting : or '{' or '=' [.]`

I could not locate the exact code change in Strongswan that changed and caused this, but fact is that the *.conf files in /etc/strongswan.d have the same syntax as the PR suggests.

### DIFF
--- a/ipsec/ovs-monitor-ipsec.in
+++ b/ipsec/ovs-monitor-ipsec.in
@@ -145,10 +145,18 @@ class StrongSwanHelper(object):
     """This class does StrongSwan specific configurations."""
 
     STRONGSWAN_CONF = """%s
-charon.plugins.kernel-netlink.set_proto_port_transport_sa = yes
-charon.plugins.kernel-netlink.xfrm_ack_expires = 10
-charon.load_modular = yes
-charon.plugins.gcm.load = yes
+charon {
+  plugins {
+    kernel-netlink {
+      set_proto_port_transport_sa = yes
+      xfrm_ack_expires = 10
+    }
+    gcm {
+      load = yes
+    }
+  }
+  load_modular = yes
+}
 """ % (FILE_HEADER)
 
     CONF_HEADER = """%s


ballock:
The checks that failed previously were some test artifacts, probably network glitch.

I have rebased the change on top of master, though.

markdgray:
The documentation will also need to be updated at https://github.com/openvswitch/ovs/blob/master/Documentation/tutorials/ipsec.rst#requirements

markdgray:
> Strongswan seems to have .opt files in the source tree with the dotted option syntax. It seems that up until version 5.6, the syntax was also accepted by Strongswan.
> 
> However, the .opt files are converted to .conf files during Strongswan build, and the dotted syntax is no longer accepted by Strongswan (tested on 5.8.2).
> 
> The effect was that the ovs ipsec monitor fails to start Strongswan, since that complains with:
> `/etc/strongswan.d/ovs.conf:4: syntax error, unexpected ., expecting : or '{' or '=' [.]`
> 
> I could not locate the exact code change in Strongswan that changed and caused this, but fact is that the *.conf files in /etc/strongswan.d have the same syntax as the PR suggests.

You could add some of this text into the commit message and add a "Signed-off-by".

ballock:
> The documentation will also need to be updated at https://github.com/openvswitch/ovs/blob/master/Documentation/tutorials/ipsec.rst#requirements and probably NEWS. What does this mean in terms of backwards compatibility?

Previous versions of Strongswan accepted the conf file syntax, no problem. I tested it with 5.6.

I just checked version 5.3.5 source code, and it does the very same thing by compiling the .opt files to .conf files and I am pretty sure the updated config will work there as well.

As for the docs, I never tested Libreswan, but this change goes in the StrongSwanHelper class, "This class does StrongSwan specific configurations.", so this change does not affect Libreswan compatibility. No reason to update the wiki for this commit.

jianjuns:
I made the same fix as this PR (in PR #334 which I just closed after I found this one). Also to confirm I tested the fix with strongSwan 5.3.5 (the version with Ubuntu 16.04.2) and saw no problem.

igsilya:
Thanks, everyone!
Patch applied to master as commit b424becaac58d8cb08fb19ea839be6807d3ed57f.  Also, this is not technically a bugfix, but it seems important to support strongswan 5.7+ as it's already really old, so I backported the fix down to 2.13.

PR #319, opened 2020-06-15
Fix Windows incompatibility when setting new feature

ruicao93:
OVS_DP_ATTR_NAME field is required when sending OVS_DP_CMD_SET to
windows kernel driver. The function "dpif_netlink_set_features"
dose not set the OVS_DP_ATTR_NAME field which will cause set feature
failure and ovs-vswitchd will exist.

This patch fixes the issue by setting "request.name" in request.

Fix https://github.com/openvswitch/ovs-issues/issues/187

Signed-off-by: Rui Cao <rcao@vmware.com>


blp:
Applied to master and branch-2.13.

PR #324, opened 2020-07-20
datapath-windows: Reset ct_mark/ct_label to support ALG

commandgjj:
The ct_mark/ct_label setting on related connection keep the same
behavior with Linux datapath. If one CT entry has parent/master entry,
its ct_mark and ct_label should inherit from the corresponding part
of parent/master entry at initialization.

Signed-off-by: Jinjun Gao <jinjung@vmware.com>


commandgjj:
The patch has merged into master. So close this PR.

PR #325, opened 2020-07-21
Additional parameters for physical DPDK ports

HofmT:
Hello developers,

**Problem**: Right now, only physical DPDK ports beginning with 'dpdk' prefix are supported in the Red Hat network scripts.
However, openvswitch no longer supports this naming pattern. See this error message from ovs 2.11.1:
```shell
error: "'dpdk-eth1-1' is missing 'options:dpdk-devargs'. The old 'dpdk<port_id>' names are not supported"
```

The current way of adding a physical DPDK port is:
```shell
ovs-vsctl add-port <br-name> <port-name> \
 -- set Interface <port-name> type=dpdk options:dpdk-devargs=<PCI_address> 
```
[See OvS doc](http://docs.openvswitch.org/en/latest/topics/dpdk/phy/)

**Suggestion**: Include a new variable specifying the pci address of a physical DPDK device.


HofmT:
@blp I saw that you originally worked on that file. Could you please comment?

HofmT:
Would appreciate if anyone could comment on this. Thank you!

PR #326, opened 2020-07-27
datapath-windows: Update flow key in SET action

commandgjj:
The flow key is not updated when process OVS_ACTION_ATTR_SET action.
It will impact follow-up actions, such as, conntrack module cannot
find created conntrack entry if passing old flow key to it.

Reported-by: Rui Cao <rcao@vmware.com>
Signed-off-by: Jinjun Gao <jinjung@vmware.com>


aserdean:
Applied via ML: https://github.com/openvswitch/ovs/commit/ba73001b6e1d1048df149d7d024e5fecd8ca618d

PR #330, opened 2020-08-03
releases doc: Meter action does not really work in upstream Linux kernels older than 4.18

reith:
Since 92d0d515d6 Open vSwitch disables meter action in linux kernel
path for 4.15, 4.16 and 4.17 but docs says Meter action is supported in
4.15 which can be misleading.

Since Ubuntu 18.04 kernel is 4.15, this can confuse many users.


horms:
Hi,

Thanks for your contribution.

PR #332, opened 2020-08-13
netdev-linux.c: Fix NULL deref reported by Coverity.

MiterV1:
When sock is NULL, calling nl_sock_listen_all_nsid triggers NULL deref.

### DIFF
--- a/lib/netdev-linux.c
+++ b/lib/netdev-linux.c
@@ -634,7 +634,9 @@ netdev_linux_notify_sock(void)
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


aserdean:
Thank you for the PR.
Sorry it took so long but it seems like the issue was fixed with the following commit:
https://github.com/openvswitch/ovs/commit/849a40ccfb9c7c6bba635b517caac4f12ab63eee

MiterV1:
yes, commit 849a40c fix it.

PR #334, opened 2020-09-11
Fix ovs-monitor-ipsec support for strongSwan >= version 5.7.0

jianjuns:
Starting from version 5.7.0, strongSwan no more supports specifying a
configuration parameter with the path delimited by dots in a
configuration file. This commit fixes the strongSwan configuration
parameters that ovs-monitor-ipsec writes, to comply with the new
strongSwan format.

### DIFF
--- a/ipsec/ovs-monitor-ipsec.in
+++ b/ipsec/ovs-monitor-ipsec.in
@@ -145,10 +145,18 @@ class StrongSwanHelper(object):
     """This class does StrongSwan specific configurations."""
 
     STRONGSWAN_CONF = """%s
-charon.plugins.kernel-netlink.set_proto_port_transport_sa = yes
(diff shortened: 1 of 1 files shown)


jianjuns:
I found out I should send the patch to dev@openvswitch.org. Will do that after I figure out how.

igsilya:
Hi. Thanks for the patch, but there is already similar patch proposed here: https://github.com/openvswitch/ovs/pull/315
It'll be great if you can help with review/testing.

BTW, we do accept pull requests, but, yes, mail-list is a primary way of submitting patches right now.

jianjuns:
Ok. PR #315 makes exactly the same change with mine. So let me close this one.
Thanks for the letting me know!

PR #337, opened 2020-09-22
Fix memory leak while parsing lldp packets

kokjo:
Remember to free previously allocated buffer with overwriting.
Normal lldp packets should not trigger these types of situations.

But specially crafted packets with multiple for instance `LLDP_TLV_PORT_ID` tlv's will leak memory.


igsilya:
Posting the link to my reply on mail-list here for consistency:
https://mail.openvswitch.org/pipermail/ovs-dev/2020-September/375457.html

vincentbernat:
Looks good! Thanks!

PR #342, opened 2020-11-20
rhel:  Change unwind-devel package

chetaiyong:
Change BuildRequires unwind-devel to libunwind-devel
There is no unwind-devel package, only libunwind-devel package is found, and no error is reported with libunwind-devel during compilation


chetaiyong:
There is no unwind-devel package, only libunwind-devel package is found, and no error is reported during compilation

chetaiyong:
> Change BuildRequires unwind-devel to libunwind-devel

PR #344, opened 2021-01-12
LLDP: fix wrong "const char*" issue which leads to the chassis descriptor always as the same

RickZhong8282:
fix wrong "const char*" issue which leads to the chassis descriptor always as the same

### DIFF
--- a/lib/ovs-lldp.c
+++ b/lib/ovs-lldp.c
@@ -203,7 +203,7 @@ aa_print_element_status_port(struct ds *ds, struct lldpd_hardware *hw)
                    &system_id_null,
                    sizeof port->p_element.system_id)) {
             const char *none_str = "<None>";
-            const char *descr = NULL;
+            char *descr = NULL;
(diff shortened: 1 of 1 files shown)


RickZhong8282:
In case multiple LLDP neighbors exist, the Rx LLDP chassis descriptors shown by the command are always the first neighbor's due to wrongly use of the "const". From the logic of the codes, the chassis ID and descriptor should be different from each neighbor. Thus the "const" should be removed.

igsilya:
`const` is just a compiler hint to show that we're not going to modify the data using this pointer. It doesn't have any runtime effects on the application logic.  We're not modifying the data using this pointer, so, `const` here is perfectly correct.
My best guess is that you had `static` keyword in mind.  But `const` is not `static`.

RickZhong8282:
Thanks for the reply. You are right. I mixtured "const" and "static" here. Let me  do more testing and see what's the root cause for my issue.

PR #348, opened 2021-02-19
Python: Support open tcp socket with domain name

lucheng0127:
When try to connect to a remote tcp socket, remote address don't
support use domain name.
Fixed it by reslove it in ovs.socket_util.inet_open_active.

### DIFF
--- a/python/ovs/socket_util.py
+++ b/python/ovs/socket_util.py
@@ -219,18 +219,16 @@ def inet_parse_active(target, default_port):
         host_name = address[0]
     if not host_name:
         raise ValueError("%s: bad peer name format" % target)
+    try:
+        # If host_name is domain name convert it to ip address string
+        host_name = socket.gethostbyname(host_name)
+    except socket.gaierror as e:
+        return get_exception_errno(e), None
     return (host_name, port)
 
 
 def inet_open_active(style, target, default_port, dscp):
     address = inet_parse_active(target, default_port)
-    try:
-        # If address is domain name convert it to ip address string
-        address_list = list(address)
-        address_list[0] = socket.gethostbyname(address_list[0])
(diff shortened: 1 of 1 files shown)


blp:
This will cause daemons to hang for arbitrary amounts of time as they try to resolve DNS names, so I don't think it's a good idea to merge it as is.

jumpojoy:
@blp thanks for review. This is up to end user to decide if DNS is a weak point in system or not. On some companies maybe it is, in other not. For example in ipv6 world DNS is mandatory nobody remember whole addresses, in kubernetes world DNS is mandatory. It seems that we may update `inet_open_active` function and replace it with `socket.create_connection` method from socket library [0] which detects address family and do all needed translations. We can put concerns about DNS availability to the documentation, but this should not be a blocker to provide ability to use hostnames. 

[0] https://docs.python.org/3/library/socket.html#socket.create_connection

blp:
@jumpojoy You didn't address the point I made in my review.

The right thing to do would be to use an asychronous DNS resolver.

igsilya:
One way to achieve the goal is to use libunbound's python library.  We use libunbound for the C version of that code and it seems to support asynchronous mode with python as well.

BubaVV:
@blp Could we assume that python-ovs runs with eventlet by default? Eventlet monkey patches socket resolver to make it asynchronous. And we get slow synchronous resolve as fallback.

Critical issue of this patch that it breaks IPv6 compatibility - it requires getaddrinfo function and some logic to parse address

igsilya:
@BubaVV implementation using linunbound was proposed recently: https://patchwork.ozlabs.org/project/openvswitch/patch/20230518035015.813298-1-twilson@redhat.com/

PR #349, opened 2021-03-22
add new command to show LLDP neighbor info

RickZhong8282:


blp:
Thanks for submitting a patch.

Please read the instructions for writing commit messages and submitting patches in Documentation/internals/contributing/submitting-patches.rst and then submit again. It's probably better to send your patch through email, since that's the main way that OVS is developed.

RickZhong8282:
Got it. Thanks for your reply.

PR #350, opened 2021-03-22
RFC - util: fallback when locked memory allocation fails

javacruft:
Normally when OVS runs on a server the effective system limit for
locked memory is unlimited as the ovs-vswitchd runs as the root
user in the root namespace of the server.

When OVS is run in a non-priviledged container a system limit will
apply and its possible that this limit will be reached during
normal operation.

If this occurs unlock all memory and re-attempt (re)allocation of
memory rather than fail and abort.

If this subsequent attempt also fails abort the process as before.

Signed-off-by: James Page <james.page@ubuntu.com>

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -41,6 +41,9 @@
 #ifdef _WIN32
 #include <shlwapi.h>
 #endif
+#ifdef HAVE_MLOCKALL
+#include <sys/mman.h>
+#endif
 
 VLOG_DEFINE_THIS_MODULE(util);
 
@@ -103,6 +106,12 @@ set_memory_locked(void)
     is_memory_locked = true;
 }
 
+void
+set_memory_unlocked(void)
+{
+    is_memory_locked = false;
+}
+
 bool
 memory_locked(void)
 {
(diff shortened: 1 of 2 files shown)


blp:
I wonder whether it makes sense to use mlockall() at all in a case like this.

If it does, then I think that the repetitive code and #if's here should be factored out into a function. It could be used something like this:
```
if (retry_after_out_of_memory())
    return xmalloc(size);
```

Naming the function is the hardest part.

javacruft:
I did consider just disabling all memory locking in ovs-lib but this did not feel like the right solution - there may be sufficient resource limits for OVS to operate with memory locking successfully inside of a container.

javacruft:
I went with your suggestion of 'retry_after_out_of_memory' - thanks for the feedback!

blp:
The fallback for xrealloc() will lose the original pointer.

igsilya:
DPDK uses memory locking for mempool pages and it relies on these pages not being moved or unloaded, because it uses physical addresses to configure hardware NICs.  Also, some DPDK drivers are using rte_mem_lock_page() -> mlock() to lock in memory pages that are used for driver-specific device I/O.  Unlocking any of these pages will likely crash the system by NMI if such a page will be moved or swapped without notifying the device and driver.

I'd say it's much more safe to just run OVS without passing `--mlockall`.

blp:
I saw a few gaps here and opportunities for improvement, so I converted it to a 3-patch series that I posted starting here: https://mail.openvswitch.org/pipermail/ovs-dev/2021-March/381494.html. I'd appreciate feedback.

PR #351, opened 2021-03-24
debian: Add configure option to debian/rules

shahar-klein:
Before this fix, in order to build with jemalloc,
I had to hijack DATAPATH_CONFIGURE_OPTS:
debian/rules -e DATAPATH_CONFIGURE_OPTS="LIBS=-ljemalloc"
Now it is possible to:
debian/rules -e EXTRA_CONFIGURE_OPTS="LIBS=-ljemalloc"

Signed-off-by: Shahar Klein <sklein@nvidia.com>


blp:
Thanks, applied to master.

PR #353, opened 2021-03-24
Add a check that either the --all or records argument provided to destroy command

roytman:
`ovn-nbctl` or `ovn-sbctl` CLI utilities allow destroying the entire table or some records from the given table.
However, either the `--all` option or the deleted records should be provided. 
This patch adds the test.

Signed-off-by: Alexey Roytman <roytman@il.ibm.com>


roytman:
The patch was submitted by email, so I'm closing this PR.

PR #357, opened 2021-04-26
fix coredump when using "ovs-ofctl add-groups"

bBobxx:
desc: When using ovs-ofctl add-groups with only "switch" argument, a
coredump is generated. The main reason is that the command "ovs-ofctl
add-groups" need two arguments but the limitation of min-args of this
command is set to 1.


bBobxx:
Looks like BSD CI is broken.

igsilya:
Hi.  FreeBSD CI is ok.  It fails because your branch is 2 years old and tries to use old base image.  Please, rebase.
Also, please, provide a Sign-off for your patch.
```

Q: What's a Signed-off-by and how do I provide one?

    A: Free and open source software projects usually require a contributor to
    provide some assurance that they're entitled to contribute the code that
    they provide.  Some projects, for example, do this with a Contributor
    License Agreement (CLA) or a copyright assignment that is signed on paper
    or electronically.

    For this purpose, Open vSwitch has adopted something called the Developer's
    Certificate of Origin (DCO), which is also used by the Linux kernel and
    originated there.  Informally stated, agreeing to the DCO is the
    developer's way of attesting that a particular commit that they are
    contributing is one that they are allowed to contribute.  You should visit
    https://developercertificate.org/ to read the full statement of the DCO,
    which is less than 200 words long.

    To certify compliance with the Developer's Certificate of Origin for a
    particular commit, just add the following line to the end of your commit
    message, properly substituting your name and email address:

        Signed-off-by: Firstname Lastname <email@example.org>

    Git has special support for adding a Signed-off-by line to a commit
    message: when you run "git commit", just add the -s option, as in "git
    commit -s".  If you use the "git citool" GUI for commits, you can add a
    Signed-off-by line to the commit message by pressing Control+S.  Other Git
    user interfaces may provide similar support.

```
More details on different commit tags you may find in a contribution guide:
https://docs.openvswitch.org/en/latest/internals/contributing/submitting-patches/#tags

PR #358, opened 2021-04-26
utilities: disable mlockall in containers

javacruft:
When OVS is run in a non-priviledged container a system limit will
apply and its possible that this limit will be reached during
normal operation.

To avoid this situation disable all memory locking when OVS is
executed within a container.

Signed-off-by: James Page <james.page@ubuntu.com>


igsilya:
Hi.
IIUC, after this change it will be not possible to use memory locking inside the container even if user wants to, e.g. if it's a privileged container or if user configured big enough limits.  Shouldn't it be a users' decision to lock or not to lock the memory?  There is an option ``--no-mlockall`` for that in ``ovs-ctl`` script.

horms:
Hi,

Thanks for your contribution.

PR #360, opened 2021-05-06
fix coredump when using "ovs-ofctl add-groups"

bBobxx:
desc: When using ovs-ofctl add-groups with only "switch" argument, a
coredump is generated. The main reason is that the command "ovs-ofctl
add-groups" need two arguments but the limitation of min-args of this
command is set to 1.

Signed-off-by: Wang Yibo <bobxxwang@126.com>

### DIFF
--- a/utilities/ovs-ofctl.c
+++ b/utilities/ovs-ofctl.c
@@ -5051,7 +5051,7 @@ static const struct ovs_cmdl_command all_commands[] = {
     { "add-group", "switch group",
       1, 2, ofctl_add_group, OVS_RW },
     { "add-groups", "switch file",
-      1, 2, ofctl_add_groups, OVS_RW },
+      2, 2, ofctl_add_groups, OVS_RW },
     { "mod-group", "switch group",
       1, 2, ofctl_mod_group, OVS_RW },
     { "del-groups", "switch [group]",


bBobxx:
Why my CI fails, I'm using a new fork repo to make this pull request.

igsilya:
> Why my CI fails, I'm using a new fork repo to make this pull request.

At the time of a previous pull request there was no issues in FreeBSD CI.  But now there is one.  I sent a patch to fix it: https://mail.openvswitch.org/pipermail/ovs-dev/2021-May/382702.html .

bBobxx:
> > Why my CI fails, I'm using a new fork repo to make this pull request.
> 
> At the time of a previous pull request there was no issues in FreeBSD CI. But now there is one. I sent a patch to fix it: https://mail.openvswitch.org/pipermail/ovs-dev/2021-May/382702.html .

Thanks for your reply. By the way, does my patch meet the merge requirements?

igsilya:
It's OK in general.  One thing is that commit `Author` field doesn't match with the `Signed-off-by` tag.  They should be the same.
```
$ git show 785f42ee8
commit 785f42ee8ae7e75d8f4b46eb61ec35d1005c7068
Author:     w00574186 <wangyibo***.com>
AuthorDate: Thu May 6 11:14:07 2021 +0800

    desc: When using ovs-ofctl add-groups with only "switch" argument, a
    coredump is generated. The main reason is that the command "ovs-ofctl
    add-groups" need two arguments but the limitation of min-args of this
    command is set to 1.
    
    Signed-off-by: Wang Yibo <bobxxwang***.com>
```

And a few nitpicks: Commit itself has no subject line, so it doesn't look pretty in the output of `git log --oneline`.  There should be a short subject line, empty line, and the rest of a description after that.  It's also a good idea to include a `Fixes` tag to the commit message to point to the commit that introduced the issue.  So, in the end, to be perfect, commit message should like something like this:
```
$ git show 785f42ee8
commit 785f42ee8ae7e75d8f4b46eb61ec35d1005c7068
Author: Wang Yibo <bobxxwang***.com>
Date:   Thu May 6 11:14:07 2021 +0800

    ovs-ofctl: Fix coredump when using "add-groups" command.
    
    When using ovs-ofctl add-groups with only "switch" argument, a coredump
    is generated.  The main reason is that the command "ovs-ofctl add-groups"
    need two arguments but the limitation of min-args of this command is
    set to 1.
    
    Fixes: 7395c05254df ("Implement OpenFlow 1.1+ "groups" protocol.")
    Signed-off-by: Wang Yibo <bobxxwang***.com>
```

bBobxx:
OK, modified as required.

igsilya:
> OK, modified as required.

Thanks! I'll get back to this patch once CI is fixed.

igsilya:
Ci should be fixed now.  If you'll rebase one more time on the latest master, all tests should pass.

igsilya:
Applied as commit f173527ec8a42e79e0a322ebe4bed999ea9ba497 and backported down to 2.12.
Thanks!

PR #361, opened 2021-05-21
[PATCH] dpif-netdev.c : Fix indentation.

MiterV1:
Add extra space to fix indentation.

Signed-off-by: miter <linhuang@ruijie.com.cn>


igsilya:
Indentation is correct here.  See reply on the mail list: https://mail.openvswitch.org/pipermail/ovs-dev/2021-May/383282.html

PR #364, opened 2021-08-08
openflow: allow to configure ofp_desc fields on bridge other_config

miguelbf-alb:
## Description

This pull request allows the user to configure the missing (currently static/forced) fields returned by the switch as a part of reply to a `OFPMP_DESC` request (ofp_desc), namely:

- **mfr_desc**: Manufacturer description (currently always Nicira, Inc.)
- **hw_desc**: Hardware description (currently always "Open vSwitch")
- **sw_desc**: Software description (currently always the version of OVS)

on the other-config of the Bridge table. 

serial_desc and dp_desc were already available, so I decided to group all of them on a single function (`bridge_configure_ofp_desc`) for the sake of simplicity.

## Motivation

SDN controllers such as [ONOS](https://github.com/opennetworkinglab/onos) make use of the information provided from the switch in the reply to `OFPMP_DESC` to load SDN controller drivers that define and control the pipeline (i.e. the sequence of tables and what match fields/actions are available in certain tables). Most of these drivers (if not all) are compatible with openvswitch. Some of them, for instance [ofdpa-ovs](https://github.com/opennetworkinglab/onos/blob/2b4de873e4033b7973b399d25cb8828a73bc2e24/drivers/default/src/main/resources/onos-drivers.xml#L162-L167) are even designed to emulate physical pipelines (ofdpa) on top of the OVS dataplane.
It'd be nice that we could configure those fields so that the appropriate driver is loaded automatically when ovs attempts to connect to the controller. 

## Tests

This was runtime tested using the latest ONOS version. In OVS the following commands were executed:

```
ovs-vsctl add-br br0
ovs-vsctl set bridge br0 other_config:mfr-desc=ONF
ovs-vsctl set bridge br0 other_config:hw-desc="OFDPA OVS"
ovs-vsctl set bridge br0 other_config:sw-desc="OFDPA OVS"
ovs-vsctl set-controller br0 tcp:xxx.xxx.xxx.xxx:6633
```

The controller automatically loaded the ovs-ofdpa driver instead of the default one.

Please kindly provide feedback
Regards

### DIFF
--- a/ofproto/ofproto.c
+++ b/ofproto/ofproto.c
@@ -797,6 +797,27 @@ ofproto_set_threads(int n_handlers_, int n_revalidators_)
 }
 
 void
+ofproto_set_mfr_desc(struct ofproto *p, const char *mfr_desc)
+{
+    free(p->mfr_desc);
+    p->mfr_desc = nullable_xstrdup(mfr_desc);
+}
+
+void
+ofproto_set_hw_desc(struct ofproto *p, const char* hw_desc)
+{
+    free(p->hw_desc);
+    p->hw_desc = nullable_xstrdup(hw_desc);
+}
+
+void
+ofproto_set_sw_desc(struct ofproto *p, const char* sw_desc)
+{
+    free(p->sw_desc);
+    p->sw_desc = nullable_xstrdup(sw_desc);
+}
+
+void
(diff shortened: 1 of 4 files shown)


miguelbf-alb:
Sent to the mailing list in https://mail.openvswitch.org/pipermail/ovs-dev/2021-August/386588.html

PR #377, opened 2022-01-05
ignore l4_hash when recv form LOCAL port

Grim-lock:


Grim-lock:
ovs recv tcp stream form LOCAL port, linux call _skb_set_hash_from_sk_ to set skb->hash and skb->l4_hash, ovs use skb->hash to select group bucket.
but when tcp retransmission occurs，linux call _sk_rethink_txhash_ to rehash skb->hash of retransmission packet, so the retransmission packet will select a different bucket than the original package.
ovs should ignore l4_hash，and call ___skb_get_hash_ to get hash

PR #379, opened 2022-01-21
Update cache line size on PowerPC

noloader:
PPC64 machines have a 128-byte L1 cache line size. See http://lists.llvm.org/pipermail/llvm-dev/2017-March/110982.html.

Signed-off-by: Jeffrey Walton noloader@gmail.com


igsilya:
It's been some time, but I applied the change now.  Sorry for delay.

PR #380, opened 2022-01-21
Add PPC64 and s390x testing to Travis

noloader:
This commit adds PPC64 and s390x testing to Travis.

Signed-off-by: Jeffrey Walton <noloader@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,19 +35,31 @@ matrix:
   include:
     - arch: arm64
       compiler: gcc
+      name: Arm64, GCC, default kernel
       env: TESTSUITE=1 DPDK=1
     - arch: arm64
       compiler: gcc
+      name: Arm64, GCC, 5.5 and 4.19 kernels
       env: KERNEL_LIST="5.5 4.19"
     - arch: arm64
       compiler: gcc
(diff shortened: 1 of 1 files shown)


igsilya:
Hi, Jeffrey.  Thanks for the patch!

The one thing here though is that since Travis CI practically dropped support of OSS projects, we had to switch to GitHub Actions and remove the Travis integration from the main repository. We left the yml file for the time being just in case someone else has a Travis account and uses it for testing.

At this point Travis CI is not part of our actual CI and therefore no-one is tracking if it still works.  This means that adding new architectures will not help us to avoid their build breakages.

If you want to have PPC64 and s390x arches added to the yml, so you personally can test them periodically, that is OK, and I can accept this patch if you promise that it works. :)  Otherwise, it doesn't have a big value for the project as we don't have a way to actually run these checks.

noloader:
Thanks @igsilya,

> If you want to have PPC64 and s390x arches added to the yml, so you personally can test them periodically, that is OK, and I can accept this patch if you promise that it works.

Ah, sorry to hear that. Travis got me too. I lost testing on about 6 open source projects. I got so fed up with Travis I had Travis delete my accounts and I removed access to my GitHub.

So yeah, it is probably better to close the request.

PR #382, opened 2022-02-23
netdev-offload: fix deadlock when adding flow table in offload condition

zhaozhanxu:
Bug reason:
When a revalidator thread of ovs tries to get the flow table,
it will nest call two read locks, namely netdev_ports_flow_get and netdev_ifindex_to_odp_port.
When other threads call the write lock between the two read locks,
there will be a deadlock problem, such as when calling netdev_ports_insert.

Fix method:
When getting the flow table, the two nested locks actually lock different tables,
but the same lock is used again, so just two different tables need to use different locks.

Bug exists in all versions since 2.13.

Signed-off-by: zhaozhanxu <zhaozhanxu@163.com>

### DIFF
--- a/lib/netdev-offload.c
+++ b/lib/netdev-offload.c
@@ -409,10 +409,11 @@ netdev_set_hw_info(struct netdev *netdev, int type, int val)
 
 /* Protects below port hashmaps. */
 static struct ovs_rwlock netdev_hmap_rwlock = OVS_RWLOCK_INITIALIZER;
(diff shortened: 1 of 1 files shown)


horms:
Hi,

Thanks for your contribution.

The change proposed by this commit appears to have been included upstream in cd608cf96eb93ebc4aa44d1393b9cb00bfde46e5.

Accordingly, I am closing this PR.

PR #389, opened 2022-06-22
vlog: fix startup fail when /var/log is full

coolMarlon:
vlog: fix startup fail when /var/log is full

When constrct /var/log full, ovsdb-server and ovs-vswitchd can not startup.
The main reasion is that OVS exit immediately after open log-file failed.

Signed-off-by: Shengwang Shi <shishengwang@huawei.com>

### DIFF
--- a/include/openvswitch/vlog.h
+++ b/include/openvswitch/vlog.h
@@ -30,6 +30,7 @@
 #include <stdarg.h>
 #include <stdbool.h>
 #include <time.h>
+#include <errno.h>
 #include <openvswitch/compiler.h>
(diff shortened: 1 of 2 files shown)


coolMarlon:
`/var/log` may full because other abnormal process.  At this monment, if ovs crash, then it can not restart successfully. 
I think this problem can be optimized. If OVS open log file, it continues to start. When /var/log has free space, OVS restore logging.

coolMarlon:
Can someone tells me why my CI fails?

horms:
Hi,

Thanks for your contribution.

PR #392, opened 2022-07-13
Fix python idl.Row.__str__ method

higherorderfunctor:
Fixes `idl.Row.__str__` to only add columns that exist on the object.  The `Row` object passed to the `updates` argument of `Idl.notify` only contains a subset of columns.  Printing that `Row` causes an `AttributeError` if there are any unchanged columns.

Signed-off-by: Christopher Aubut [christopher@aubut.me](mailto:christopher@aubut.me)


igsilya:
@otherwiseguy, could you, please take a look at this one?

otherwiseguy (approved):
Looks good to me. Thanks!

PR #402, opened 2022-10-30
lib/byteq: add unit tests for better code coverage

yuvalif:
Signed-off-by: Yuval Lifshitz <ylifshit@redhat.com>


yuvalif:
no interest in the PR

chaudron:
Hi @yuvalif

The common workflow for OVS is through email, so it would be better to submit your patch patches through email. This way more contributors will see it and this will leverage various automation through patchwork, and you will get the robot to send you messages when some tests fail.

For more details on the patch submission guidelines see https://github.com/openvswitch/ovs/blob/master/Documentation/internals/contributing/submitting-patches.rst

Cheers,

Eelco

PR #404, opened 2022-11-09
Do not remove offloaded flows for running netdev

gotostack:
In ret_flow offloading environment, we met a strange behavior. When deleting a netdev port, the offloaded flows of other ports (with running streams) will be unexpectedly removed. And new offloaded flows were not added anymore.

So let's gdb it, when delete a netdev port, we have a call stack like this: -> port_del
-> dpif_netdev_port_del
-> do_del_port
-> dp_netdev_offload_flush

In this ``dp_netdev_offload_flush`` function, it will call ``dp_netdev_offload_flush_enqueue`` to add DP_OFFLOAD_FLUSH items to the offloading queue. Then this thread hits the barrier, and switching to the offloading thread.

The offloading thread will do the following works: -> dp_netdev_flow_offload_main
-> dp_offload_flush
-> netdev_flow_flush
-> netdev_flow_api.flow_flush(netdev_offload_dpdk_flow_flush) 
It will run the following code:

        if (data->netdev != netdev && data->physdev != netdev) {
            continue;
        }
        if (data->creation_tid == tid) {
            netdev_offload_dpdk_flow_destroy(data);
        }

Because offloading can be only available to one physdev, offloaded items have flows belong to other ports, so ``netdev_offload_dpdk_flow_destroy`` will remove those offloaded flows for the running ports. Added a INFO log before the ``netdev_offload_dpdk_flow_destroy`` which proves these behaviors.

        if (data->creation_tid == tid) {
            VLOG_INFO("Remove offloaded dpdk flows for netdev %s",
                      netdev_get_name(netdev));
            netdev_offload_dpdk_flow_destroy(data);
        }

Logs are like this:
    INFO|Remove offloaded dpdk flows for netdev vdpb519e956-17
    INFO|Remove offloaded dpdk flows for netdev vdpb519e956-17
    INFO|Remove offloaded dpdk flows for netdev vdpb519e956-17
    INFO|Remove offloaded dpdk flows for netdev vdpb519e956-17
    INFO|Remove offloaded dpdk flows for netdev ens10f0
    INFO|Remove offloaded dpdk flows for netdev ens10f0
    INFO|Remove offloaded dpdk flows for netdev ens10f0
    INFO|Remove offloaded dpdk flows for netdev ens10f0

The LOG count is equal to the offloaded flows of 4 UDP streams (4 for ingress, 4 for egress).

This patch adds a ``is_deleting`` attribute to the netdev.hw_info. It is initialized to false. In the ``dp_netdev_offload_flush``, it will be set to true. And in the ``netdev_offload_dpdk_flow_destroy``, it will check if the netdev->hw_info.is_deleting.
Then, the running ports' flows will be remained.

Signed-off-by: LIU Yulong <i@liuyulong.me>

### DIFF
--- a/lib/dpif-netdev.c
+++ b/lib/dpif-netdev.c
@@ -3101,6 +3101,7 @@ dp_netdev_offload_flush(struct dp_netdev *dp,
     ovs_barrier_init(&barrier, 1 + netdev_offload_thread_nb());
 
     netdev = netdev_ref(port->netdev);
+    netdev->hw_info.is_deleting = true;
     dp_netdev_offload_flush_enqueue(dp, netdev, &barrier);
     ovs_barrier_block(&barrier);
     netdev_close(netdev);
--- a/lib/netdev-offload-dpdk.c
+++ b/lib/netdev-offload-dpdk.c
@@ -2546,7 +2546,9 @@ netdev_offload_dpdk_flow_flush(struct netdev *netdev)
         if (data->netdev != netdev && data->physdev != netdev) {
             continue;
         }
-        if (data->creation_tid == tid) {
+        if (data->creation_tid == tid && data->netdev->hw_info.is_deleting) {
+            VLOG_INFO("Remove offloaded dpdk flows for netdev %s",
+                      netdev_get_name(netdev));
             netdev_offload_dpdk_flow_destroy(data);
         }
     }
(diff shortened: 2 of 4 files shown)


gotostack:
Not sure if this is the best way to fix the problem, since this problem may caused by other code, so we may also have solutions like:
1. let ``dp_netdev_offload_flush_enqueue `` only gather the deleting port's flush items.
2. let the offloading thread re-offload the flows for the running netdev ports (may be there is a deadlock).
3. the thread_id before calling the ``netdev_offload_dpdk_flow_destroy`` is not right. fix it. And remove flows for the deleting dev only.

PR #410, opened 2023-03-09
emc_insert: fix the bug of wrong replacement

0x777a6c:
In original implementation, when the first entry is alive, the second entry is not alive and the hash value of the second entry is less than the first, the replacement will occur on the second entry. This commit fix the problem.


chaudron:
Thanks for the patch, but it looks like your change is not passing the check patch script. For details see the following document in the repo https://github.com/openvswitch/ovs/blob/master/Documentation/internals/contributing/submitting-patches.rst. 

In addition, it would be preferred to submit patches through email, as this will leverage various automation through patchwork, and you will get the robot to send you messages when some tests fail. 

But more important this way other people working on this project will see your patch, as most of them are not monitoring the GitHub pull requests.

Cheers,

Eelco

horms:
Hi,

Thanks for your contribution.

PR #411, opened 2023-04-12
Branch 2.17

situnan:
centos7.2 build fail (Linux localhost.localdomain 3.10.0-327.el7.x86_64)

./configure -with-linux=/lib/modules/$(uname -r)/build
make

error info（kenel not support  rxhash，but ./configure check exist HAVE_RXHASH ):
/root/openvswitch-2.17.6/datapath/linux/compat/include/linux/skbuff.h: In function ‘__skb_set_hash’:
/root/openvswitch-2.17.6/datapath/linux/compat/include/linux/skbuff.h:462:5: error: ‘struct sk_buff’ has no member named ‘rxhash’
  skb->rxhash = hash;

reason:  configure shell check rxhash  exist but kernel is not support now

kernel code want use hash;not rxhash :
	RH_KABI_REPLACE(__u32	rxhash,
			__u32	hash)

configure code（check error logic：grep 'rxhash' $KSRC/include/linux/skbuff.h ）：
  { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether rxhash matches in $KSRC/include/linux/skbuff.h" >&5
$as_echo_n "checking whether rxhash matches in $KSRC/include/linux/skbuff.h... " >&6; }
  if test -f $KSRC/include/linux/skbuff.h; then
    grep 'rxhash' $KSRC/include/linux/skbuff.h >/dev/null 2>&1
    status=$?
    case $status in
      0)
        { $as_echo "$as_me:${as_lineno-$LINENO}: result: yes" >&5
$as_echo "yes" >&6; }

  echo '#define HAVE_RXHASH 1' >> datapath/linux/kcompat.h.new


igsilya:
Hi.  I'm not sure what this PR supposed to do.  Assuming it's just an attampt for a bug report.  Please, send bug reports to the ovs-discuss@openvswitch.org or create an issue in the ovs-issues repo instead.

For the described issue: It's hard to keep the code compatible with heavily modified downstream kernels like RHEL/CentOS ones.  Out-of-tree kernel module is deprecated since 2.15 and fully removed in 3.0.  CentOS 7.2 is a 7 (?) year old release and CentOS 7 as a whole is going EOL in a few months.  So, it doesn't make a lot of sense to add support now.
I'd suggest to upgrade to some more modern distribution, so you don't need to use the deprecated kernel module.

PR #420, opened 2023-12-15
tests: using ulimit to ignore coredump.

chenxy1988:
Some testcases simulating a fault with kill command, that might cause system generate a coredump file.
Using ulimit to disable coredump before that testcases and restore them after testing.

### DIFF
--- a/tests/daemon-py.at
+++ b/tests/daemon-py.at
@@ -42,7 +42,11 @@ check_ancestors $child $monitor
 
 # Kill the daemon process, making it look like a segfault,
 # and wait for a new child process to get spawned.
+AT_CHECK([ulimit -c > rlimit.core])
+AT_CHECK([ulimit -c 0])
 AT_CHECK([kill -SEGV $child])
+AT_CHECK([ulimit -c `cat rlimit.core`])
+AT_CHECK([rm rlimit.core])
 OVS_WAIT_WHILE([kill -0 $child])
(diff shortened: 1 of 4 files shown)


igsilya:
Hi @chenxy1988 .  Thanks for the patch.  Though I'm not sure we can actually do that.  First, if something fails we need to be able to restore the value.  Also, it looks like it's not even possible to even restore a value.  For example, I'm getting a permission error on Fedora 38:

```
$ ulimit -c
unlimited
$ ulimit -c 0
$ ulimit -c
0
$ ulimit -c unlimited
bash: ulimit: core file size: cannot modify limit: Operation not permitted
```
The test would just fail at this point.

chenxy1988:
> Hi @chenxy1988 . Thanks for the patch. Though I'm not sure we can actually do that. First, if something fails we need to be able to restore the value. Also, it looks like it's not even possible to even restore a value. For example, I'm getting a permission error on Fedora 38:
> 
> ```
> $ ulimit -c
> unlimited
> $ ulimit -c 0
> $ ulimit -c
> 0
> $ ulimit -c unlimited
> bash: ulimit: core file size: cannot modify limit: Operation not permitted
> ```
> 
> The test would just fail at this point.

Thanks for you information, I tested them on an embedded linux that there is only a root user and forgetting to test them on a distribution linux.

Linux allows everybody to reduce the limit, but increasing can only be done by root, so closing this PR and I have to find another way to set/restore the coredump file size.

Thanks!

PR #423, opened 2024-03-22
fix ovs ipf crash

l15310488637:
Description:

when 1) The fragment timeout is between 15 seconds and 25 seconds; 2) The ipf_list currently has received more than 32 fragments, and there are other fragments of same big packet that have not been received.

When the above two scenario conditions are met, due to exceeding the capacity of the packet batch(here is 32), ipf_dp_packet_batch_add returns false, and ipf_list will not be cleared. However, the 32 fragments packets added to the packet batch will be processed normally. When receiving the subsequent fragments of the ipf_list, because the first 32 fragments have been processed, when processing subsequent fragment packets, relevant information about the processed fragment packets will be read，therefore will occur carsh. One solution is do not forward timeout fragment packets from the above scenarios, that is, do not add them to the packet batch, and handle other scenarios according to the original logic.

### DIFF
--- a/lib/ipf.c
+++ b/lib/ipf.c
@@ -1011,7 +1011,7 @@ ipf_purge_list_check(struct ipf *ipf, struct ipf_list *ipf_list,
 }
 
 /* Does the packet batch management and common accounting work associated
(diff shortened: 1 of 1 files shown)


horms:
Hi,

Thanks for your contribution.

PR #426, opened 2024-07-28
Fix typo in openvswitch-switch.README.Debian

dqadri:
Fixed a typo describing the linux kernel module vs the 3rd party module.


igsilya:
Hi, @dqadri .

Thanks for your contribution.

I am closing this PR as patch submissions via the email workflow, as described in [Submitting Patches](https://docs.openvswitch.org/en/latest/internals/contributing/submitting-patches/), are the preferred mechanism for contributions to Open vSwitch.

Note: I believe the intended word was 'This' and not 'The', it may not even be a typo but the old English version of a word, though unlikely.  Also, this whole documentation section will need to be removed soon since OOT module is deprecated long ago, so I'm not sure if it is worth fixing.

